PR #3, opened 2026-07-02
Some basic repo setup

viktorpusTT:
* Tooling: ruff, pre-commit, CI gate
* codeowners
* Make the repo conform to the pre-commit (import ordering and newlines)

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,0 +1,3 @@
+# Order is important; the last matching pattern takes precedence.
+
+* @viktorpusTT @sanjaradylov @tchedaTT
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,20 @@
+name: ci
+
+on:
+  pull_request:
+  push:
+    branches: [dev]
+
+# Only lint runs here. The unit tests import vllm_tt_plugin, which imports the
+# locally built TT vLLM (`empty` target) that is not installable on a stock
+# GitHub runner, so pytest belongs on a TT environment (see tt-metal's
+# vllm-nightly-tests), not in this workflow.
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.10"
+      - uses: pre-commit/action@v3.0.1
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,40 @@
+# Python
+__pycache__/
+*.py[cod]
+*$py.class
+*.egg-info/
+.eggs/
+build/
+dist/
+
+# Virtual environments
+.venv/
+venv/
+env/
+
+# Test and analysis output
+.coverage
+.coverage.*
+coverage.xml
+htmlcov/
+.pytest_cache/
+.ruff_cache/
+.mypy_cache/
+.pyre/
+
+# Local environment files
+.env
+.env.*
+
+# Logs and crash dumps
+*.log
+core
+core.*
+
+# Local model, tensor, and profiler artifacts
+model_cache/
+tt_cache/
+vllm_cache/
+*.tensorbin
+tracy/
+profiler/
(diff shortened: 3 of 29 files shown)


sanjaradylov (approved):

agent-shade-reviewer (approved):
- role: reviewer
- session: `1.WPUH1cNFlcXf0MpaZypgr0yONvfLkedZDKk1qivW5r605A4lrifkMvaUXIV86pc_331zAKMDiJQ`

**Verdict: approve.** Reviewed head `1b720a6` (base `c1aca66`); the brief named `b8d7473`, and `1b720a6` sits on top of it as the scalar-xargs fix. Five files, +780/-1. Source-checked against the served vLLM 0.29.0 and the ninfer reference; host suite reported green by the worker (419 passed, 3 failures + 66 errors unchanged from base).

Lenses, in order:

1. **Greedy identity — passes.** The only changed egress condition is `_pure_greedy` (mtp.py:220-221): `thinking_token_budget is None` and `not self._post_thinking_switched(req_id)`. A no-budget, unswitched request still satisfies it (the budget processor is in `known_processors`, mtp.py:205), so device sampling and the greedy egress are untouched. The one new per-step cost is the per-row `phase.advance()` in `PostThinkingSwitch.apply` (thinking.py:414), which `refresh_logitsprocs` (input_batch.py:621-622, 1017-1018) calls every step while a reasoning parser is configured — including on rows that carry no field. <challenged> Not a defect: it is the cost of the switch being default-on, and it is O(new tokens) per row. It is the line that breaks the "no new per-step host work" guarantee on the no-feature path, and it is named here.

2. **Gating — passes.** A budget-tracked row forces `has_active_logitsprocs` true (input_batch.py:186-190 via `has_tracked_rows`) → host sampling; in MTP it rides the per-request policy (mtp.py:99-101) and `_sample` (mtp.py:145). The device sampler is untouched for untracked rows. No double enforcement: vLLM's `thinking_budget_state_holder` stays `None` (the plugin never constructs it), and `BUILTIN_LOGITS_PROCESSORS` does not include the budget processor, so the sampler does not duplicate it.

3. **Budget vs ninfer — passes with a note.** Boundary ids come from the qwen3 reasoning parser (`<think>`/`</think>`, natural_end == end). Exact hit: the model gets exactly `budget` reasoning tokens, then the forced close (thinking.py:196-201 emits the end sequence one token per step). <challenged> Two deviations from ninfer's four rules (run-ninfer.sh:125-153) are not stated in the PR: (a) the close is the parser's `</think>` end sequence, not ninfer's 25-token control suffix; (b) there is no 26-token-minimum / 400 rejection — a budget at or above the effective limit silently never fires (ninfer rejects below 26 under, and silently disengages at/above the limit). Both are defensible; the pins fragment should state them so the parity record is complete.

4. **Sampler switch — passes.** Fires once (the `_switched` set, thinking.py:412-417), only on an output-side close (`advance` consumes output only; a prompt-side open never closes), per-request fields win over the deployment preset (`_request_preset`, thinking.py:299-334), and the preset is recorded (NINFER_POST_THINKING 0.2/20/0.95/0.0, thinking.py:294-296). It writes the persistent per-row tensors (thinking.py:418-421), which every sampling path reads.

5. **Tests — passes.** 17 functions / 20 cases (the opt-out test is parametrized over the four scalar opt-out forms). All contract: boundary forcing, prompt-opened count, stop-at-close, inert no-field, rollback, row move (asserts the untracked row is untouched), row removal, install skip without reasoning config, switch-once + no re-apply, non-reasoning turn untouched, request override wins, opt-out, deployment preset, deployment off, absent without reasoning config, invalid setting raises. Deterministic, no RNG/device. <challenged> No test for a prompt with a completed prior section (opened + natural_end) — the multi-turn `preserve_thinking` case; the logic is covered by `starts_in_reasoning`.

6. **Blast radius — passes.** Five plugin files only. No serve.sh, no ninfer, no MTP speculative-config change.

Notes (challenged, non-blocking):
- The PR body is stale relative to head `1b720a6`: it says "16 cases" (now 17 functions / 20 cases) and describes the request override as an object (`vllm_xargs: {"post_thinking": {...}}`), but the head takes scalars (`post_thinking_temperature`, …) because `vllm_xargs` carries scalars only.
- The budget deviations from ninfer's four rules (lens 3) should be stated in the PR body / pins fragment.

Device evidence is still pending a serving window (per the PR body).

PR #4, opened 2026-07-03
Compatibility fixes with vLLM v0.24.0

viktorpusTT:
Compatibility fixes for running the TT plugin against upstream vLLM v0.24.0. Each commit is an independent fix for a behavior change between the old 0.16 fork and 0.24.

## Changes

- **`platform: implement TTPlatform.set_device as no-op`** - 0.24's multiproc executor calls `current_platform.set_device` from its async-output-copy thread. The base `Platform.set_device` raises `NotImplementedError`, killing that thread. TT device context is owned by the ttnn mesh device, not a torch device context, so a no-op is correct.

- **`platform: auto-fit max_model_len to TT KV cache capacity`** - 0.24's `get_kv_cache_configs` now rejects configs where the model's `max_model_len` (often the HF default, e.g. gemma-4-12B's 262144) needs more KV than TT's `num_gpu_blocks_override` provides, crashing EngineCore. TT sizes the KV cache from `max_tokens_all_users`, decoupled from per-request `max_model_len`. Opt into vLLM's native auto-fit (`original_max_model_len=-1`) to clamp `max_model_len` down to the servable length, and add the `update_max_model_len` worker hook the engine calls afterward (absent on `WorkerBase`).

- **`scheduler: keep grammar-blocked prefills out of decode batches`** - 0.24 segregates requests blocked on grammar compilation into a separate `skipped_waiting` queue. The TT prefill/decode routing only inspected `waiting`, so the base scheduler promoted a ready structured request into a running-decode step, violating the all-prefill-or-all-decode invariant and tripping the model_runner batch assert.

- **`tt-plugin: fold GPT-OSS gather-DP into lane-DP`** - GPT-OSS is served by the tt_transformers generator, which drives data parallelism inside a single process, so gathered multi-process DP is unnecessary: `--data_parallel_size` folds into in-process TT lanes like the Galaxy generators. Gated on model architecture plus `data_parallel_size > 1`; other models keep gather-DP. Config-gating is covered by host tests; the lane execution path with GPT-OSS is not yet validated on Galaxy hardware.

- **`platform: force compact JSON from structured output grammars`** - 0.24 defaults `disable_any_whitespace` to `False`, so under greedy decoding the model can loop on grammar-legal whitespace until the token budget is spent, returning truncated unparseable JSON.

- **`tests: cap sampling temperatures at 2.0 for upstream vllm compat`** - 0.24 rejects temperature outside `[0, 2]` at the OpenAI request-validation layer (400 BadRequest); the old 0.16 fork had no upper bound, so temperature 5/10/50 reached the sampler. 2.0 keeps the variety these tests assert on while staying inside the accepted range.

# CI

TBD

### DIFF
--- a/src/vllm_tt_plugin/async_decode.py
+++ b/src/vllm_tt_plugin/async_decode.py
@@ -3,6 +3,7 @@
 
 from __future__ import annotations
 
+import os
 import threading
 import time
 from dataclasses import dataclass, fields, replace
@@ -13,7 +14,16 @@ import ttnn
 from vllm.v1.outputs import AsyncModelRunnerOutput, LogprobsLists, ModelRunnerOutput
 
 from vllm_tt_plugin.input_batch import SEED_NONE_SENTINEL
+from vllm_tt_plugin.logger import init_tt_logger
 from vllm_tt_plugin.structured_output import has_structured_outputs
+
+logger = init_tt_logger(__name__)
+
+# Verbose mrope/decode-reset tracing, off by default. Enable by exporting
+# ``VLLM_TT_DEBUG_MROPE=1`` on the server process to confirm the rope-delta
+# re-application on the prefill->decode transition (see ``submit_decode``).
+_DEBUG_MROPE_ENV = os.environ.get("VLLM_TT_DEBUG_MROPE", "")
+DEBUG_MROPE = _DEBUG_MROPE_ENV not in ("", "0", "false", "False")
 
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import GrammarOutput, SchedulerOutput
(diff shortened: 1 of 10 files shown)


sanjaradylov:
**FYI**

Two major CI failures are:
* Correctness < 100% on `[WH-T3K] Qwen3-VL-32B-Instruct`: usually a couple of examples with incoherent field values ([job](https://github.com/tenstorrent/tt-metal/actions/runs/28667420669/job/87096627020)); e.g.:
   ```bash
   {"name": "n", "company": "co",
    "age": 22222222222222222222...  # Too long, reaching `max_tokens` and not closing with '}'
   ```
* Non-deterministic decoding, presence-penalty, and other constrained-sampling errors on `[WH-GLX] Qwen3-32B device with sampling tests`.

---

- Generally, both jobs are difficult to complete from the first attempt because of device errors.
- The second type of failure happens in some of the previous scheduled runs from `main` (e.g., example [1](https://github.com/tenstorrent/tt-metal/actions/runs/29060435463/job/86263516661), [2](https://github.com/tenstorrent/tt-metal/actions/runs/28985613725/job/86016949737)) albeit with different sampling-error rate. However, the [last](https://github.com/tenstorrent/tt-metal/actions/runs/29296045352/job/87284644356) run was successful.
- Tried to resolve both errors through normalization of greedy requests, but apparently these fixes were already introduced in [`tt-metal`](https://github.com/tenstorrent/tt-metal/pull/45532) for Llama Galaxy (and hence, for Qwen3-32B) and the errors still remain (e.g., [1](https://github.com/tenstorrent/tt-metal/actions/runs/29311309696/job/87039211288), [2](https://github.com/tenstorrent/tt-metal/actions/runs/29329753495/job/87280674931)).

tchedaTT:
Do we want to auto-fit max_model_len? IMO this may be misleading to users when a requested config is changed silently. We already have a patch for this.

https://github.com/tenstorrent/vllm/blob/826b15a812da6f73c40cf45de75c8230a6353346/plugins/vllm-tt-plugin/src/vllm_tt_plugin/worker.py#L68 from https://github.com/tenstorrent/vllm/pull/438 is the patch we currently have for this purpose and should be modified as we include upstream's solution when doing this rebase

viktorpusTT:
Split into separate PRs, closing.

PR #7, opened 2026-07-17
platform: auto-fit max_model_len to TT KV cache capacity

viktorpusTT:
vLLM 0.24's `get_kv_cache_configs` now rejects configs where the model's `max_model_len` (often the HF default, e.g. gemma-4-12B's 262144) needs more KV than TT's `num_gpu_blocks_override` provides, crashing EngineCore. TT sizes the KV cache from `max_tokens_all_users`, decoupled from per-request `max_model_len`. Opt into vLLM's native auto-fit (`original_max_model_len=-1`) to clamp `max_model_len` down to the servable length, and add the `update_max_model_len` worker hook the engine calls afterward (absent on `WorkerBase`).

Split out of #4 (compatibility fixes for upstream vLLM v0.24.0).

### DIFF
--- a/src/vllm_tt_plugin/platform.py
+++ b/src/vllm_tt_plugin/platform.py
@@ -495,6 +495,19 @@ class TTPlatform(Platform):
             )
             model_config.max_logprobs = MAX_TOP_K
 
+        # Opt into vLLM's auto-fit of max_model_len against the KV cache the TT
+        # device can actually hold. The TT worker sizes the KV cache from the
+        # model's total token budget (max_tokens_all_users) and pins it via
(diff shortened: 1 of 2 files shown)


sanjaradylov (approved):

tchedaTT:
https://github.com/tenstorrent/vllm/blob/826b15a812da6f73c40cf45de75c8230a6353346/plugins/vllm-tt-plugin/src/vllm_tt_plugin/worker.py#L68 from https://github.com/tenstorrent/vllm/pull/438 is the patch we currently have for this purpose and should be modified as we include upstream's solution

PR #8, opened 2026-07-17
scheduler: keep grammar-blocked prefills out of decode batches

viktorpusTT:
vLLM 0.24 segregates requests blocked on grammar compilation into a separate `skipped_waiting` queue. The TT scheduler must revisit that queue during prefill scheduling so ready requests can be promoted, and must hide both waiting queues during decode-only scheduling to prevent mixed prefill and decode batches.

PR #12 already landed the shared scheduler queue handling while replacing legacy gathered-DP with standard DP or single-process lane-DP. This PR now completes the lane-DP behavior:

- Count `skipped_waiting` when negotiating lane-wide prefill intent, so grammar-ready requests are revisited and promoted.
- Keep forced-prefill fallback coordinated by `TTLaneCoordinator`; an individual lane must not fall back to decode while another lane successfully prefills.
- Add regression coverage for lane intent, decode queue isolation, and lane-wide fallback ownership.
- Clarify the separate roles of prefill intent and decode queue isolation.

`engine.py` remains unchanged because the standalone plugin now routes through upstream standard DP or single-process lane-DP after #12. The separate gathered-DP livelock in the vLLM fork is tracked in https://github.com/tenstorrent/vllm/issues/450.

Split out of #4 (compatibility fixes for upstream vLLM v0.24.0).

### DIFF
--- a/src/vllm_tt_plugin/engine.py
+++ b/src/vllm_tt_plugin/engine.py
@@ -323,7 +323,13 @@ class TTDPEngineCoreProc(DPEngineCoreProc):
 
     def _dp_negotiate_forced_mode(self) -> TTSchedulingMode:
         has_running = bool(getattr(self.scheduler, "running", []))
-        has_waiting = bool(getattr(self.scheduler, "waiting", False))
+        # ``skipped_waiting`` holds prefill requests blocked on grammar
+        # compilation; count them so a rank with only grammar-blocked
+        # structured-output work still signals prefill intent instead of
+        # letting the base scheduler promote them into a gathered decode step.
+        has_waiting = bool(getattr(self.scheduler, "waiting", False)) or bool(
+            getattr(self.scheduler, "skipped_waiting", False)
+        )
         max_running = getattr(self.scheduler, "max_num_running_reqs", 0)
         has_capacity = len(getattr(self.scheduler, "running", [])) < max_running
         local_prefill_intent = (
--- a/src/vllm_tt_plugin/engine.py
+++ b/src/vllm_tt_plugin/engine.py
@@ -323,13 +323,7 @@ class TTDPEngineCoreProc(DPEngineCoreProc):
 
     def _dp_negotiate_forced_mode(self) -> TTSchedulingMode:
         has_running = bool(getattr(self.scheduler, "running", []))
-        # ``skipped_waiting`` holds prefill requests blocked on grammar
-        # compilation; count them so a rank with only grammar-blocked
-        # structured-output work still signals prefill intent instead of
-        # letting the base scheduler promote them into a gathered decode step.
-        has_waiting = bool(getattr(self.scheduler, "waiting", False)) or bool(
-            getattr(self.scheduler, "skipped_waiting", False)
-        )
+        has_waiting = bool(getattr(self.scheduler, "waiting", False))
         max_running = getattr(self.scheduler, "max_num_running_reqs", 0)
         has_capacity = len(getattr(self.scheduler, "running", [])) < max_running
         local_prefill_intent = (
--- a/src/vllm_tt_plugin/lane_scheduler.py
+++ b/src/vllm_tt_plugin/lane_scheduler.py
@@ -356,8 +356,14 @@ class TTLaneCoordinator(SchedulerInterface):
         A lane wants to prefill when it has queued requests and either nothing
         running (so it must prefill to make progress) or spare capacity to admit
         more alongside its running decodes.
+
+        ``skipped_waiting`` holds prefill requests blocked on grammar
+        compilation and counts as queued work: without it a lane whose only
+        pending request is a grammar-blocked structured-output request would
+        signal decode-only, and the base scheduler could then promote that
+        request into a mixed prefill+decode step.
         """
-        has_waiting = bool(sched.waiting)
+        has_waiting = bool(sched.waiting) or bool(sched.skipped_waiting)
         has_running = bool(sched.running)
         has_capacity = len(sched.running) < self._per_lane_max
         return int(has_waiting and ((not has_running) or has_capacity))
--- a/src/vllm_tt_plugin/lane_scheduler.py
+++ b/src/vllm_tt_plugin/lane_scheduler.py
@@ -358,7 +358,9 @@ class TTLaneCoordinator(SchedulerInterface):
         more alongside its running decodes.
 
         ``skipped_waiting`` holds prefill requests blocked on grammar
-        compilation. We want to try to schedule them, because otherwise base scheduler won't revisit and promote them - decode intent hides that queue.
+        compilation. We want to try to schedule them, because otherwise the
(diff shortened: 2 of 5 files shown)


src/vllm_tt_plugin/lane_scheduler.py
```diff
@@ -356,8 +356,14 @@ def _local_prefill_intent(self, sched: TTScheduler) -> int:
         A lane wants to prefill when it has queued requests and either nothing
         running (so it must prefill to make progress) or spare capacity to admit
         more alongside its running decodes.
+
+        ``skipped_waiting`` holds prefill requests blocked on grammar
+        compilation and counts as queued work: without it a lane whose only
+        pending request is a grammar-blocked structured-output request would
+        signal decode-only, and the base scheduler could then promote that
+        request into a mixed prefill+decode step.
```
tchedaTT: This comment is slightly misleading - base scheduler not adding skipped_waiting to a decode batch relies on us hiding skipped_waiting when scheduling decodes, not on skipped_waiting implying prefill intent

src/vllm_tt_plugin/engine.py
```diff
@@ -323,7 +323,13 @@ def _dp_any_rank_has_scheduler_requests(self) -> bool:
 
     def _dp_negotiate_forced_mode(self) -> TTSchedulingMode:
         has_running = bool(getattr(self.scheduler, "running", []))
-        has_waiting = bool(getattr(self.scheduler, "waiting", False))
+        # ``skipped_waiting`` holds prefill requests blocked on grammar
+        # compilation; count them so a rank with only grammar-blocked
+        # structured-output work still signals prefill intent instead of
+        # letting the base scheduler promote them into a gathered decode step.
```
tchedaTT: Same as other comment - this is not the actual correctness requirement.
The reason we need skipped_waiting to imply prefill intent is actually because the promotion from skipped_waiting to waiting happens only in the base scheduler when scheduling prefills.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -65,8 +73,17 @@ def __init__(self, *args, **kwargs):
     def set_forced_mode(self, mode: TTSchedulingMode) -> None:
         self._forced_mode = mode
 
+    def _has_pending_prefill(self) -> bool:
+        """Whether any request is waiting to be prefilled.
+
+        Includes ``skipped_waiting`` so a request blocked on grammar
+        compilation still routes scheduling through the prefill-only path
+        instead of leaking into a decode step.
```
tchedaTT: this isn't the part that prevents prefill leaking into decode

src/vllm_tt_plugin/scheduler.py
tchedaTT, line 117: I think we need a similar fallback in the forced-mode path, or else we will unnecessarily spin when grammar compiles in the following scenario:
I we have 31 requests in decode and one in skipped_waiting, we will try to schedule a prefill. If that request's grammar hasn't compiled yet, we will schedule zero tokens. This fallback here only gets triggered when we don't force mode, so it won't be triggered - we will do the same again, until the prefill actually gets scheduled.
I think this isn't actually specific to skipped_waiting, but applies to any scenario where we have prefill intent but schedule zero tokens.
Lane scheduler has a fallback, so this only gets triggered in the gathered-dp scenario.

Actually, I think this is even worse.
If zero tokens are scheduled not due to skipped_waiting, but due to eg. no kv cache capacity, this will deadlock (livelock) and keep spinning forever, because no decodes will be scheduled to clear it. 

This is preexisting. Please check if what I wrote makes sense and file an issue if it does. I think this also could affect multihost deepseek and might need to be fixed in the other repo.

src/vllm_tt_plugin/engine.py
viktorpusTT: Removed the `engine.py` change from this PR. The standalone plugin legacy gathered-DP path is being deprecated and will be removed soon, so no fix is needed there. The corresponding issue for the vLLM fork path that will survive longer is tracked in tenstorrent/vllm#450.

src/vllm_tt_plugin/scheduler.py
viktorpusTT: Confirmed the zero-token forced-prefill livelock. Lane-DP already performs a coordinated fallback after merging all lane outputs. The standalone plugin legacy gathered-DP path is being deprecated and removed soon, so this PR does not add a fallback there. The vLLM fork path is tracked separately in tenstorrent/vllm#450: https://github.com/tenstorrent/vllm/issues/450.

src/vllm_tt_plugin/scheduler.py
viktorpusTT: Updated the docstring. The helper causes default scheduling to revisit blocked prefills and ensures forced decode enters `_schedule_decode_only()` when only `skipped_waiting` is populated. Hiding both queues inside `_schedule_decode_only()` is the actual batch-isolation mechanism.

src/vllm_tt_plugin/lane_scheduler.py
viktorpusTT: Updated the wording to separate the two requirements. Counting `skipped_waiting` as lane prefill intent lets the base prefill loop revisit and promote the request once its grammar is ready. Hiding `skipped_waiting` in `_schedule_decode_only()` is what prevents admission into a decode batch.

viktorpusTT:
Rebased onto `main` after #12 merged.

The shared `skipped_waiting` handling from #12 is preserved. This PR now adds the missing lane intent handling and removes the per-scheduler forced-prefill fallback introduced by #12. Forced mode is lane-wide, so only `TTLaneCoordinator` may switch all lanes to decode after confirming that the merged prefill step scheduled zero tokens. An individual lane falling back locally could otherwise create a mixed prefill and decode batch.

The focused diff now touches only the lane scheduler, shared scheduler, and their tests. `engine.py` remains unchanged.

src/vllm_tt_plugin/lane_scheduler.py
```diff
@@ -356,8 +356,15 @@ def _local_prefill_intent(self, sched: TTScheduler) -> int:
         A lane wants to prefill when it has queued requests and either nothing
         running (so it must prefill to make progress) or spare capacity to admit
         more alongside its running decodes.
+
+        ``skipped_waiting`` holds prefill requests blocked on grammar
+        compilation and counts as queued work: without it a lane whose only
+        pending request is a grammar-blocked structured-output request would
+        signal decode-only. Decode-only scheduling hides that queue, so the
+        base scheduler would never revisit it to promote the request after its
+        grammar becomes ready.
```
tchedaTT:
```suggestion
        ``skipped_waiting`` holds prefill requests blocked on grammar
        compilation. We want to try to schedule them, because otherwise base scheduler won't revisit and promote them - decode intent hides that queue.
```

tchedaTT (approved):

PR #12, opened 2026-07-17
[feature] Port standard-DP runtime to standalone plugin:

sanjaradylov:
## Purpose

Ports the standalone-plugin-compatible subset of [tenstorrent/vllm#430](https://github.com/tenstorrent/vllm/pull/430) into the current plugin branch. The fork will keep legacy DP for multi-host models, so this branch takes the standard-DP behavior that can live entirely in the standalone plugin.

## Modifications

- For DP > 1, non-Galaxy/non-GPT-OSS models now follow upstream vLLM standard DP instead of the legacy TT gathered path.
- Galaxy-generator and GPT-OSS models still use single-process lane mode. We transparently convert `--data_parallel_size N` into `N` in-process TT lanes in **platform.py** (via `_model_folds_gather_dp_into_lanes`).
- The legacy user config `tt_data_parallel_size` no longer drives parallelism. If present, it is ignored; effective TT lane count is derived internally; see **config.py**.
- In standard DP, every rank owns its own mesh, loads the model, allocates KV cache, and owns its host sampler state; see **worker.py**.
- Single-host standard DP no longer needs the TT MPI launcher just to assign chips. We now discover per-rank TT chip groups from the real TT submesh split, expose those groups through `TTPlatform.device_id_to_physical_device_id`, and let the upstream `CoreEngineLauncher` / `CoreEngineProcManager` assign `TT_VISIBLE_DEVICES` per rank; see **platform.py**.
- That single-host device-group discovery runs in a short-lived spawned helper process, so the main server process does not hold TT chip locks before worker startup; see **platform.py**.
- The TT launcher path is now reserved for explicit MPI / rank-binding / multi-node placement. If `rank_binding`, `mpi_args`, `nnodes > 1`, or `node_rank > 0` is provided, we still use `TTCoreEngineLauncher`; see **platform.py**.
- When the explicit TT launcher path is used, one MPI rank must still match one DP rank, and every `rank_binding` entry must provide `TT_VISIBLE_DEVICES`; see **launcher.py**.
- Rank-local mesh sizing is constrained by `TT_VISIBLE_DEVICES` instead of assuming the full-machine `MESH_DEVICE` preset. We also skip fabric setup when the rank-local mesh has only one device; see **worker.py**.
- `TTScheduler` fixes: removed stale `throttle_prefills` forwarding (base `Scheduler.schedule()` signature mismatch), added `skipped_waiting` handling to prevent grammar-blocked requests leaking into decode batches (port of PR #8's fix), and added a livelock fallback for forced `PREFILL_ONLY` mode.
- Model runner: converted the hard mixed-batch assertion into a defensive runtime guard that filters decode requests from an unexpected prefill step instead of crashing the engine.
- Host-side regression coverage for standard-DP routing, device discovery, and launcher validation lives in **test_dp_modes.py**.

## CI branch

Standard DP is evaluated via tt-metal nightly CI on branch [`sanjaradylov/vllm-nightly-oot-plugin`](https://github.com/tenstorrent/tt-metal/compare/main...sanjaradylov/vllm-nightly-oot-plugin), which references this standalone plugin instead of the fork's embedded copy and introduces supporting workflow fixes.

## Test Plan

1. Unit tests covering DP class selection, rank mapping, and other related configs.
2. Test DP=1 for any regressions locally. Pay attention to structured-output tests after recent execute-sample changes from `dev`.
3. Run DP=1 nightly CI jobs.
4. Run local DP > 1 on T3K.
5. Run DP > 1 CI jobs via `sanjaradylov/vllm-nightly-oot-plugin`.

## Test Result

- [x] Unit tests passing (including new test_dp_modes.py).
- [x] DP=1 (P150, blackhole) pass locally for Qwen.
- [x] DP=1 actions pass for [NoOp ](https://github.com/tenstorrent/tt-metal/actions/runs/28355444919)and [Llama](https://github.com/tenstorrent/tt-metal/actions/runs/28358105906/job/84007473821), and [Gemma 3](https://github.com/tenstorrent/tt-metal/actions/runs/28369151364/job/84048343763) and [4](https://github.com/tenstorrent/tt-metal/actions/runs/28501089215).
- [x] DP > 1 CI via [`sanjaradylov/vllm-nightly-oot-plugin`](https://github.com/tenstorrent/tt-metal/compare/main...sanjaradylov/vllm-nightly-oot-plugin) [passing](https://github.com/tenstorrent/tt-metal/actions/runs/30241546191).


src/vllm_tt_plugin/worker.py
```diff
@@ -65,6 +66,50 @@ class CompilationTimes(NamedTuple):
 register_tt_models(register_test_models=_should_pre_register_tt_test_models_from_cli())
 
 
+def _rank_owns_mesh(parallel_config: Any) -> bool:
```
viktorpusTT: When can this be False?

src/vllm_tt_plugin/platform.py
```diff
@@ -634,6 +980,23 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
         else:
             vllm_config.scheduler_config.scheduler_cls = TT_SCHEDULER_CLS
 
+        parallel_config.dp_engine_core_proc_cls = "vllm.v1.engine.core.DPEngineCoreProc"
```
viktorpusTT, line 1014: Is this read anywhere?

viktorpusTT (approved):
Approve if CI shows no new bugs. File perf regressions (if any) as new issues.

tchedaTT:
I think a few issues might have snuck in here - please take a look at this AI summary

[P1] The mixed-batch “guard” corrupts scheduler/runner agreement
The guard identifies non-resumed cached requests as running/decode requests and removes their row indices while building model inputs. [model_runner.py#L903-L934](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L903-L934)
But by then:
The scheduler has emitted a SchedulerOutput containing those requests.
vLLM has advanced request.num_computed_tokens for every scheduled request. [scheduler.py#L1130-L1176](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/core/sched/scheduler.py#L1130-L1176)
With async scheduling, it has also added output placeholders. [async_scheduler.py#L19-L44](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/core/sched/async_scheduler.py#L19-L44)
_update_states() has applied the complete scheduler output to the persistent input batch before _prepare_model_inputs() performs filtering. [model_runner.py#L598-L699](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L598-L699)
The guard updates none of that. It does not:
remove the requests from SchedulerOutput;
roll back computed-token accounting or async placeholders;
release/adjust KV allocations;
preserve a filtered request-ID-to-output-row mapping.
The comment saying “they will be re-scheduled next” is therefore unjustified. The scheduler already believes they executed.
The partial-filter path is also likely to crash later because _build_runner_output() defaults to the full persistent batch size and asserts the device output row count matches it. [model_runner.py#L2835-L2856](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L2835-L2856)
Recommendation: remove the filter, fix the scheduler invariant, and retain a descriptive RuntimeError/assertion as invariant detection. Recovery would have to happen before scheduler bookkeeping is committed, not inside the model-input builder.
[P1] Forced-prefill fallback can itself create a mixed lane batch
In forced PREFILL_ONLY mode, each individual TTScheduler locally falls back to decode when its prefill schedules zero tokens. [scheduler.py#L95-L104](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/scheduler.py#L95-L104)
That is unsafe with TTLaneCoordinator:
Lane A cannot admit its prefill and locally schedules decode.
Lane B successfully schedules prefill.
The coordinator merges both outputs.
Its global fallback does not run because the merged token count is nonzero.
The result is exactly the mixed prefill/decode batch the hardware supposedly forbids.
This is confirmed by the repository’s immediate follow-up commit, which removes local fallback so all lanes switch modes together: [b7c8b59 — “keep forced fallback lane-wide”](https://github.com/tenstorrent/vllm-tt-plugin/commit/b7c8b591ad9fff1b73725d68b562f69f5af1d483).
Recommendation: apply that fix and its regression test.
[P1] Grammar-blocked lane requests can stall indefinitely
TTLaneCoordinator._local_prefill_intent() checks only sched.waiting, not sched.skipped_waiting. [lane_scheduler.py#L353-L374](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/lane_scheduler.py#L353-L374)
If a lane’s only pending request is waiting for grammar compilation:
The request resides in skipped_waiting.
The coordinator chooses decode-only.
Decode-only scheduling deliberately hides skipped_waiting.
The base scheduler never revisits the request after its grammar becomes ready.
That is an indefinite stall, including after other running requests finish.
Again, the repository’s follow-up commit confirms the defect and changes the coordinator to inspect both queues: [c020492 — grammar-blocked prefill fix](https://github.com/tenstorrent/vllm-tt-plugin/commit/c020492b7a1afb67cbd1fbf2bc2ad4cd00d55d4b).
[P1] Inherited TT_VISIBLE_DEVICES defeats per-rank isolation
_ensure_visible_devices_env() returns immediately whenever TT_VISIBLE_DEVICES is already nonempty. [worker.py#L75-L102](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/worker.py#L75-L102)
That breaks the common case where the parent process already has a visibility restriction. Stock vLLM 0.24 no longer temporarily patches the device-control environment per engine process; it calculates and stores assigned_physical_gpu_ids instead. [vLLM engine utils](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/engine/utils.py#L180-L210)
Consequently, every engine child can inherit the same parent TT_VISIBLE_DEVICES. The recovery helper sees it, returns, and never applies the discovered per-rank group. Multiple DP ranks may then attempt to open the same chips/full mesh.
Recommendation: when stored standard-DP groups exist, select the group using data_parallel_index and enforce that value even if the environment is already set. Do this before TT runtime initialization. Explicit MPI remains unaffected because it has no stored discovery groups and should preserve its rank-binding environment.
[P2] The key DP regression test targets a fork-only vLLM API
The added test calls engine_utils.get_device_indices(). [test_dp_modes.py#L441-L459](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/tests/test_dp_modes.py#L441-L459)
That function does not exist in the plugin’s pinned stock vllm==0.24.0; v0.24 uses get_physical_gpu_ids_for_local_dp_rank(). [vLLM implementation](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/engine/utils.py#L283-L346)
So this test was apparently copied from tenstorrent/vllm#430 without being ported to the actual standalone dependency. If executed under the documented environment, it fails with AttributeError and, more importantly, tests the old environment-assignment behavior rather than the actual v0.24 path.
[P3] The updated documentation still describes gathered DP as current
The README first correctly says standard DP is independent per-rank execution, then says other model families still use gathered DP. [README.md#L308-L316](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/README.md#L308-L316), [README.md#L347-L353](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/README.md#L347-L353)
docs/SCHEDULING.md similarly declares gathered DP historical/fork-only, but several current-flow sections still describe gathered scheduling, coordination, and gather/scatter as active plugin behavior. [SCHEDULING.md#L263-L286](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/docs/SCHEDULING.md#L263-L286), [SCHEDULING.md#L320-L360](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/docs/SCHEDULING.md#L320-L360)
That is particularly misleading in a PR whose main purpose is replacing gathered DP.

viktorpusTT:
Thank you. Created as https://github.com/tenstorrent/vllm-tt-plugin/issues/15

PR #14, opened 2026-07-27
platform: preserve explicit max_model_len, auto-fit only when omitted

viktorpusTT:
## Purpose

Follow-up to #7, addressing @tchedaTT's review feedback there: `_validate_tt_kv_cache_capacity` (from tenstorrent/vllm#438, ported here in #5) "should be modified as we include upstream's solution."

Two problems on `main` today, both introduced by adopting upstream's solution alongside the older stopgap rather than in place of it:

1. **`check_and_update_config` sets `original_max_model_len = -1` unconditionally**, so vLLM auto-fits *every* configuration. An explicit `--max-model-len 200000` that the TT KV pool cannot serve is silently clamped down instead of being rejected, and the user is never told the value they asked for is not in effect.
2. **`_validate_tt_kv_cache_capacity` is now dead weight.** It exists because vLLM used to validate `max_model_len` against profiled *bytes*, which the in-tree plugin faked as `1 << 64`, so the real block-count limit went unchecked — the #436 livelock. Both halves of that gap are now closed: since #12 the worker returns a byte budget that divides straight back into the TT block count, and vllm-project/vllm#41069 rewrites `available_memory` to `num_gpu_blocks_override * _pool_bytes_per_block(groups)` *before* both auto-fit and the admission check, which resolves to the same number wherever the override is visible in-process. Upstream therefore plans against the real TT pool from either side. With auto-fit forced on, the TT check can no longer fire at all; where it could, it pre-empts a better upstream error that reports an estimated servable length.

## Background

TT does not profile device memory. `get_num_available_blocks_tt()` derives a block count from the model's `max_tokens_all_users` — a per-model, per-device token budget declared by the model class in tt-metal (default 131072, sometimes an env override such as `GEMMA4_MAX_TOKENS_ALL_USERS`), plus headroom for vLLM's worst-case block allocation and for sliding-window groups on hybrid models. `TTWorker.determine_available_memory()` publishes that count both as `cache_config.num_gpu_blocks_override` and as an equivalent byte budget; the latter is what reaches the engine-side KV planner, which under the upstream multiprocess executor runs in a different process than the worker that set the override.

That pool is the budget shared by **all** concurrent requests, and it is independent of `max_model_len`, the **per-request** context limit, which normally comes from the HF config (e.g. gemma-4-12B's 262144). The two have to be reconciled somewhere; the question this PR settles is *by whom, and how loudly*.

## Change

Let vLLM's parsed `original_max_model_len` stay authoritative, with one narrow exception.

- **Explicit values are never rewritten.** A numeric value reaches upstream's override-aware capacity check unchanged; an explicit `-1` stays `-1`.
- **An omitted value falls back to auto-fit, with a warning.** Preserving `None` as well would be the purer policy, but it would abort startup for every model whose HF-derived default exceeds the TT pool — including the default path of `examples/offline_inference_tt.py`, which passes `max_model_len=None`. The warning states what the user is getting, because auto-fit stops at the length that fills the *whole* pool: maximum concurrency `1.00x`, i.e. one request can occupy the entire KV cache and stall the rest.
- **`_validate_tt_kv_cache_capacity` is removed**, along with its `get_max_concurrency_for_kv_cache_config` import and its call in `initialize_from_config`.
- **`TTWorker.update_max_model_len` stays.** `WorkerBase` defines no such hook (only `gpu_worker` does), and the engine RPCs it whenever auto-fit reduced the value, so without it the RPC raises `AttributeError`. Its comment is corrected to cover both auto-fit entry points.
- The `max_model_len` contract is documented in the README under *Configuration*.

## Resulting behavior

| `--max-model-len` | `original_max_model_len` | Result |
| --- | --- | --- |
| `8192` (fits) | `8192` | Preserved. Upstream admission check passes against the override-clamped pool. No RPC. Concurrency ≈ `pool / 8192`. |
| `200000` (does not fit) | `200000` | Preserved, then upstream aborts in `get_kv_cache_configs`: *"To serve at least one request with the model's max seq len (200000) … the estimated maximum model length is N."* Before device allocation. **Silently clamped on `main` today** — this is the regression the PR fixes. |
| `-1` | `-1` | Auto-fit reduces to the pool, `collective_rpc("update_max_model_len")` syncs the workers, and `EngineCoreReadyResponse.max_model_len` syncs the API-server process. |
| omitted | `None` → `-1` | Same as `-1`, plus the warning. Preserves today's behavior for launch commands and examples that pass no flag. |

`update_max_model_len` fires only in the bottom two rows. In lane mode (uniproc executor) the worker shares the same `ModelConfig` object, so the assignment is redundant but harmless; in standard DP the workers are separate processes and it is required. `TTModelRunner` reads `self.model_config.max_model_len` live when it builds the persistent input batch in `initialize_kv_cache`, which the engine calls *after* the RPC, so no stale copy survives. `TTCoreEngineLauncher` overrides only `prepare_launch`, `get_engines_to_handshake`, and `launch_remote_engines`, so the front-end ready-response sync is inherited intact in standard-DP mode.

## Notes

- **The `max_num_batched_tokens` bump is unaffected but slightly stale on the auto-fit path.** #438's bump and `verify_max_model_len` run in `check_and_update_config`, using the pre-fit `max_model_len`. Auto-fit only ever *lowers* `max_model_len`, so the `max_num_batched_tokens >= max_model_len` invariant still holds; the only cost is host block-table buffers sized for the larger value. Left alone deliberately — re-deriving scheduler limits from the worker would desynchronize the engine-core and worker copies. Documented in a comment.
- **No guard is needed for `max_model_len`-dependent pool budgets.** `get_num_available_blocks_tt` passes `max_model_len` into `get_max_tokens_all_users`, and DeepSeek-R1-0528 on Wormhole returns `max_model_len * max_num_seqs`, so pool sizing runs before auto-fit with the pre-fit value. That is safe: the pool is then `max_num_seqs` times what one request needs, and `_auto_fit_max_model_len` never expands beyond the current value, so it does not reduce. The combination "pool depends on `max_model_len`" **and** "auto-fit reduces" is unreachable for the current model set.
- **Nothing here applies to `tenstorrent/vllm`.** That fork is on upstream v0.16.0 (merged Apr 22 2026), which predates both vllm-project/vllm#41069 (Apr 28 2026) and the `EngineCoreReadyResponse` front-end sync. `_validate_tt_kv_cache_capacity` must stay there until it rebases past 0.24.

## Test plan

- [x] `pre-commit run` on the staged files — all hooks pass
- [x] `python3 -m py_compile` on the three changed Python files
- [ ] `pytest tests --ignore=tests/tt -v` — **not runnable locally**: this macOS host has Python 3.9 and no `ttnn`, `vllm`, or `pytest`, and `tests/test_dp_modes.py` imports `ttnn` at module scope. CI required before merge.
- [ ] Hardware smoke test on each row of the table above, in particular that an oversized explicit `--max-model-len` now aborts with upstream's estimated-length message rather than starting up clamped.

New host tests in `tests/test_dp_modes.py`:

- `test_check_and_update_config_preserves_explicit_max_model_len` (`8192`, `-1`)
- `test_check_and_update_config_auto_fits_omitted_max_model_len` (`None` → `-1`)
- `test_update_max_model_len_syncs_worker_model_config`


src/vllm_tt_plugin/platform.py
```diff
@@ -842,18 +842,46 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+        # Explicit values -- numeric or ``-1`` -- are therefore left untouched.
+        # Only an omitted value falls back to auto-fit, because the HF-derived
+        # default (e.g. gemma-4-12B's 262144) almost always exceeds the TT pool
+        # and would otherwise abort startup for every such model.
+        #
+        # Auto-fit runs after the chunked-prefill bump above, which already
+        # raised ``max_num_batched_tokens`` to the pre-fit ``max_model_len``.
+        # Auto-fit only ever lowers ``max_model_len``, so the
+        # ``max_num_batched_tokens >= max_model_len`` invariant still holds; the
+        # only cost is host block-table buffers sized for the larger value.
```
tchedaTT: This comment is quite verbose

src/vllm_tt_plugin/platform.py
```diff
@@ -842,18 +842,46 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+        if model_config.original_max_model_len is None:
+            logger.warning(
+                "max_model_len was not set; fitting it to the TT KV cache "
+                "capacity. Auto-fit stops at the length that fills the whole "
+                "pool, i.e. maximum concurrency 1.00x, so one request can "
+                "occupy the entire KV cache. For serving, pass --max-model-len "
+                "explicitly (roughly max_tokens_all_users / max_num_seqs)."
+            )
+            model_config.original_max_model_len = -1
```
tchedaTT: Why do we want to do auto-fit when max_model_len isn't set? It seems to me that if we didn't change it here, it would behave like the fork does, which matches upstream

src/vllm_tt_plugin/platform.py
```diff
@@ -842,18 +842,9 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
-        # vLLM's _check_enough_kv_cache_memory raises when max_model_len needs
-        # more KV than the override provides. Setting original_max_model_len=-1
-        # makes get_kv_cache_configs run _auto_fit_max_model_len, which clamps
-        # max_model_len down to the largest length the override-backed capacity
-        # can serve (it never expands, so a smaller user-set max_model_len is
-        # preserved) and syncs the reduced value to workers.
-        model_config.original_max_model_len = -1
+        # Preserve vLLM's max_model_len policy: omitted uses the HF-derived
+        # value, -1 auto-fits, and a positive value is used as given. The TT
+        # worker reports the real KV capacity for upstream's capacity check.
```
tchedaTT:
```suggestion
```
stranded comment

tests/test_dp_modes.py
```diff
@@ -103,6 +104,47 @@ def register_dummy_model(
...
+        assert vllm_config.model_config.original_max_model_len is None
+
+    def test_update_max_model_len_syncs_worker_model_config(self) -> None:
+        worker_instance = TTWorker.__new__(TTWorker)
+        worker_instance.model_config = SimpleNamespace(max_model_len=262_144)
+
+        TTWorker.update_max_model_len(worker_instance, 131_072)
+
+        assert worker_instance.model_config.max_model_len == 131_072
+
```
tchedaTT: given we don't have a codepath that would do it anymore, maybe unnecessary?

tchedaTT: also, do we run these tests somewhere?

tchedaTT (approved):

tests/test_dp_modes.py
viktorpusTT: Two tests folded into one, worker updating can still happen.
As for the unit tests in general: https://github.com/tenstorrent/vllm-tt-plugin/issues/19

PR #27, opened 2026-08-06
[fix] Bind 'TT_VISIBLE_DEVICES' to rank's group; make DP tests up-to-date.

### LINKED ISSUE #15: Potential issues in true-DP code
Posted by @tchedaTT in https://github.com/tenstorrent/vllm-tt-plugin/pull/12#issuecomment-5100706054:

[P1] The mixed-batch “guard” corrupts scheduler/runner agreement
The guard identifies non-resumed cached requests as running/decode requests and removes their row indices while building model inputs. [model_runner.py#L903-L934](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L903-L934)
But by then:
The scheduler has emitted a SchedulerOutput containing those requests.
vLLM has advanced request.num_computed_tokens for every scheduled request. [scheduler.py#L1130-L1176](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/core/sched/scheduler.py#L1130-L1176)
With async scheduling, it has also added output placeholders. [async_scheduler.py#L19-L44](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/core/sched/async_scheduler.py#L19-L44)
_update_states() has applied the complete scheduler output to the persistent input batch before _prepare_model_inputs() performs filtering. [model_runner.py#L598-L699](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L598-L699)
The guard updates none of that. It does not:
remove the requests from SchedulerOutput;
roll back computed-token accounting or async placeholders;
release/adjust KV allocations;
preserve a filtered request-ID-to-output-row mapping.
The comment saying “they will be re-scheduled next” is therefore unjustified. The scheduler already believes they executed.
The partial-filter path is also likely to crash later because _build_runner_output() defaults to the full persistent batch size and asserts the device output row count matches it. [model_runner.py#L2835-L2856](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/model_runner.py#L2835-L2856)
Recommendation: remove the filter, fix the scheduler invariant, and retain a descriptive RuntimeError/assertion as invariant detection. Recovery would have to happen before scheduler bookkeeping is committed, not inside the model-input builder.
[P1] Forced-prefill fallback can itself create a mixed lane batch
In forced PREFILL_ONLY mode, each individual TTScheduler locally falls back to decode when its prefill schedules zero tokens. [scheduler.py#L95-L104](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/scheduler.py#L95-L104)
That is unsafe with TTLaneCoordinator:
Lane A cannot admit its prefill and locally schedules decode.
Lane B successfully schedules prefill.
The coordinator merges both outputs.
Its global fallback does not run because the merged token count is nonzero.
The result is exactly the mixed prefill/decode batch the hardware supposedly forbids.
This is confirmed by the repository’s immediate follow-up commit, which removes local fallback so all lanes switch modes together: [b7c8b59 — “keep forced fallback lane-wide”](https://github.com/tenstorrent/vllm-tt-plugin/commit/b7c8b591ad9fff1b73725d68b562f69f5af1d483).
Recommendation: apply that fix and its regression test.
[P1] Grammar-blocked lane requests can stall indefinitely
TTLaneCoordinator._local_prefill_intent() checks only sched.waiting, not sched.skipped_waiting. [lane_scheduler.py#L353-L374](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/lane_scheduler.py#L353-L374)
If a lane’s only pending request is waiting for grammar compilation:
The request resides in skipped_waiting.
The coordinator chooses decode-only.
Decode-only scheduling deliberately hides skipped_waiting.
The base scheduler never revisits the request after its grammar becomes ready.
That is an indefinite stall, including after other running requests finish.
Again, the repository’s follow-up commit confirms the defect and changes the coordinator to inspect both queues: [c020492 — grammar-blocked prefill fix](https://github.com/tenstorrent/vllm-tt-plugin/commit/c020492b7a1afb67cbd1fbf2bc2ad4cd00d55d4b).
[P1] Inherited TT_VISIBLE_DEVICES defeats per-rank isolation
_ensure_visible_devices_env() returns immediately whenever TT_VISIBLE_DEVICES is already nonempty. [worker.py#L75-L102](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/src/vllm_tt_plugin/worker.py#L75-L102)
That breaks the common case where the parent process already has a visibility restriction. Stock vLLM 0.24 no longer temporarily patches the device-control environment per engine process; it calculates and stores assigned_physical_gpu_ids instead. [vLLM engine utils](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/engine/utils.py#L180-L210)
Consequently, every engine child can inherit the same parent TT_VISIBLE_DEVICES. The recovery helper sees it, returns, and never applies the discovered per-rank group. Multiple DP ranks may then attempt to open the same chips/full mesh.
Recommendation: when stored standard-DP groups exist, select the group using data_parallel_index and enforce that value even if the environment is already set. Do this before TT runtime initialization. Explicit MPI remains unaffected because it has no stored discovery groups and should preserve its rank-binding environment.
[P2] The key DP regression test targets a fork-only vLLM API
The added test calls engine_utils.get_device_indices(). [test_dp_modes.py#L441-L459](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/tests/test_dp_modes.py#L441-L459)
That function does not exist in the plugin’s pinned stock vllm==0.24.0; v0.24 uses get_physical_gpu_ids_for_local_dp_rank(). [vLLM implementation](https://github.com/vllm-project/vllm/blob/ee0da84ab9e04ac7610e28580af62c365e898389/vllm/v1/engine/utils.py#L283-L346)
So this test was apparently copied from https://github.com/tenstorrent/vllm/pull/430 without being ported to the actual standalone dependency. If executed under the documented environment, it fails with AttributeError and, more importantly, tests the old environment-assignment behavior rather than the actual v0.24 path.
[P3] The updated documentation still describes gathered DP as current
The README first correctly says standard DP is independent per-rank execution, then says other model families still use gathered DP. [README.md#L308-L316](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/README.md#L308-L316), [README.md#L347-L353](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/README.md#L347-L353)
docs/SCHEDULING.md similarly declares gathered DP historical/fork-only, but several current-flow sections still describe gathered scheduling, coordination, and gather/scatter as active plugin behavior. [SCHEDULING.md#L263-L286](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/docs/SCHEDULING.md#L263-L286), [SCHEDULING.md#L320-L360](https://github.com/tenstorrent/vllm-tt-plugin/blob/6942ae88e978e4fee118525af150064abda17d41/docs/SCHEDULING.md#L320-L360)
That is particularly misleading in a PR whose main purpose is replacing gathered DP.

sanjaradylov:
Resolves #15 (device-related issues)

## Purpose

In standard DP, vLLM does not rewrite the device-control env var per engine-core process, so every rank inherits the launching process's `TT_VISIBLE_DEVICES` verbatim. The previous worker logic returned early whenever the env var was already set, which left all ranks pointing at the same chips.

## Modifications

- Renamed `_ensure_visible_devices_env` to `_bind_visible_devices_env`: the inherited value is now overwritten from this rank's discovered device group instead of being trusted; see **worker.py**.
- Explicit MPI launches bind their own ranks and store no groups, so they keep their inherited value unchanged.
- A missing group for a rank now raises instead of silently falling through to the inherited value.
- Binding moved to the top of `init_device()`, before `check_and_update_config` and model registration, since model-module resolution can touch the TT runtime.
- Regrouped DP-related tests and add tests for device visibility; problematic cases are up-to-date with vLLM v0.24.

## Test Plan

1. Unit tests for per-rank binding and the missing-group error path.
2. Run DP = 1 locally for regressions.
3. Run DP > 1 on T3K and verify each rank binds a distinct chip group.

## Test Result

- [x] Unit tests passing.
- [x] DP = 1 passes locally.
- [x] DP > 1 ranks bind distinct device groups.
- [x] CI [results](https://github.com/tenstorrent/tt-metal/actions/runs/31602160497).


viktorpusTT:
The diagnosis is right and the direction is right. Overwriting the inherited value instead of
trusting it is clearly correct, and moving the bind ahead of `check_and_update_config` is the
right call. Most of what follows is about the two places where the implementation still trusts
something it should not, plus a batch of comments and test-side issues.

I have treated the MPI / rank-binding launcher path as out of scope for the standalone plugin,
so I am not calling the `xfail`ed launcher tests a production breakage. That changes how they
should be handled, though, see item 7.

---

## Blocking

### 1. `getattr(parallel_config, "data_parallel_index", 0)` silently reintroduces the bug

`src/vllm_tt_plugin/worker.py:92`

```python
dp_index = getattr(parallel_config, "data_parallel_index", 0)
```

If that attribute is ever missing, every rank binds `groups[0]`: all ranks on the same chips,
which is exactly the failure this PR exists to remove, only now with a log line saying "Bound"
to reassure you it worked.

`data_parallel_index` is a mandatory `Field(init=False)` on `ParallelConfig`, unconditionally
assigned in `__post_init__`, and the plugin sets it itself in `platform.py`. It is also exactly
the kind of internal field that moved during the 0.24 bump that motivated this PR, which is the
reason a silent default here is the wrong instinct.

The error policy inside this function is also inverted. It raises loudly for
`dp_index >= len(groups)`, which discovery already made unreachable (`utils/dp_discovery.py`
raises when `len(submeshes) != data_parallel_size`, and groups are only stored from that
result), and defaults silently for the case that can actually misfire.

Please read the attribute directly and let it raise. If you keep the range check, make it
two-sided (`not 0 <= dp_index < len(groups)`), because today a negative index quietly selects
`groups[-1]`.

### 2. The comment justifying the `init_device` reorder describes a mechanism that does not exist

`src/vllm_tt_plugin/worker.py:209`

> Bind the device group before anything can touch the TT runtime: model registration below
> resolves and imports TT model modules.

Three problems:

- There is no model registration below. `register_tt_models(...)` runs at module import time
  at `worker.py:71`, before any worker instance exists.
- Registration does not import anything. `platform.py:482` says so explicitly: it is a lazy
  `"module:Class"` string that vLLM resolves later.
- "Before anything can touch the TT runtime" is not literally true. `import ttnn` at
  `worker.py:12` already ran in this process.

The actual dependency is `TTPlatform.check_and_update_config` below, which calls
`get_model_architecture`, and that is what imports the tt-metal model module.

The invariant worth writing down is the stronger one: UMD latches `TT_VISIBLE_DEVICES` at the
first PCIe enumeration / cluster-descriptor construction, so the very first cluster touch
anywhere in this process's import graph fixes the visible set permanently. One future
import-time `ttnn.cluster.*` call in the plugin's import graph turns this rebind into a silent
no-op with no error anywhere. That is what the comment should say, because it is what a future
reader needs in order to not break this.

This comment is the only record of why the call was hoisted, so as written it sends the next
person after a hazard that is not there.

### 3. The rank's device group is re-derived from a private copy, in a different index space, when vLLM already hands it to the worker

`src/vllm_tt_plugin/worker.py:92`, `src/vllm_tt_plugin/platform.py:730`

vLLM 0.24 calls `set_assigned_physical_gpu_ids_for_dp_rank(vllm_config, local_dp_rank, ...)`
per engine core for non-CUDA/non-XPU platforms. That resolves through
`TTPlatform.device_id_to_physical_device_id` and lands this rank's group on
`parallel_config.assigned_physical_gpu_ids`, which is pickled into the worker and republished
there by the multiproc executor. The correct value is already sitting on the config the worker
holds.

`_bind_visible_devices_env` ignores it and re-derives the same thing from
`additional_config["_tt_standard_dp_visible_groups"]`, and the two index the same list
differently:

| Consumer | Index |
|---|---|
| upstream `set_assigned_physical_gpu_ids_for_dp_rank` | `local_dp_rank` |
| `init_device` mesh selection, six lines below the bind | `data_parallel_rank_local` |
| the PR's own new test at `tests/test_visible_devices.py:223` | `local_dp_rank=1` |
| `_bind_visible_devices_env` | `data_parallel_index`, i.e. the **global** DP rank |

Being fair about impact: with single-host standard DP only, `data_parallel_index` and
`data_parallel_rank_local` are equal, so this is latent rather than live today. But the PR adds
a hard `raise` on that index, which converts any future divergence from a soft fallback into a
worker-startup crash, and nothing anywhere asserts the two channels agree. Maintaining a second
source of truth for "which chips is this rank on", keyed differently from the one upstream
already populated, is the part I would rather not merge.

Suggested shape: read `self.parallel_config.assigned_physical_gpu_ids` and write that into the
env var, keeping the `additional_config` lookup only as an explicitly labelled fallback for
older vLLM. If you prefer to keep the manual lookup, at minimum index by
`data_parallel_rank_local` so it matches the line six below it, and add an assertion that the
bound value equals `assigned_physical_gpu_ids` when upstream populated it.

A related question while you are in here: `device_id_to_physical_device_id` returns
`groups[device_id]`, i.e. a string like `"16,17,18,19,20,21,22,23"`, into
`assigned_physical_gpu_ids`, which upstream types as `list[int]` and documents as consumed at
platform/topology boundaries. The new test at `tests/test_visible_devices.py:212` now asserts
that shape, which locks it in. The override has no docstring explaining that TT reinterprets
"device id" as "DP rank" and "physical device id" as "that rank's whole group". Either return
real ints and let upstream's slice build the group, or keep the string and document the
deviation on the override so the assertion is not a landmine.

---

## Please address

### 4. `platform.py` still carries the explanation this PR repudiates

`src/vllm_tt_plugin/platform.py:101`, `_store_standard_dp_visible_groups`:

> Indexed by DP rank so worker subprocesses can recover `TT_VISIBLE_DEVICES` from
> `data_parallel_index` when the env-var does not propagate through the engine-core fork chain.

That is verbatim the theory this PR declares wrong, and it sits on the producer end of the
exact payload whose consumer docstring now says the opposite. Anyone tracing the mechanism hits
both and gets contradictory root causes. It also no longer describes the code: nothing is
recovered, the value is unconditionally overwritten, and you renamed `_ensure_` to `_bind_`
specifically to kill that framing.

Suggest reducing it to the storage contract: indexed by DP rank, persisted on
`additional_config` because it has to survive pickling into the worker subprocess.

### 5. The new docstring states a version-specific fact as timeless and cites no symbol

`src/vllm_tt_plugin/worker.py:80`

> vLLM does not rewrite the device-control env var per engine-core process.

True as of 0.24, false before it. Anyone who checks against an older tree finds
`set_device_control_env_var(vllm_config, local_dp_rank)` wrapping `proc.start()`, doing exactly
what the docstring says vLLM does not do, and concludes the comment is wrong. The "(v0.24
changes)" qualifier lives only in the PR title and never reaches the code.

Please date it and name both halves, something like: as of 0.24, `CoreEngineProcManager` no
longer sets the device-control env var per engine core (`set_device_control_env_var` was
replaced by `set_assigned_physical_gpu_ids_for_dp_rank`, which writes
`parallel_config.assigned_physical_gpu_ids`); tt-metal reads only the env var, so the worker
has to materialize it here. That makes the claim checkable.

### 6. The new `RuntimeError` is undocumented and the message is not actionable

`src/vllm_tt_plugin/worker.py:93`

The function went from returning silently to aborting a worker at `init_device`, but the
docstring above still enumerates exactly two outcomes (bind, keep-inherited) and never mentions
the third, which makes the no-groups and short-groups cases read as behaving the same. They do
not.

The message itself does not say which index space `dp_index` is in (global DP rank, which is
the first thing anyone debugging it will guess wrong), does not point at the producer, and
interpolates the entire group list, which on a Galaxy DP32 run is 32 comma-separated strings on
one log line.

Also worth a line in the docstring: the `if inherited == visible_devices: return` short circuit
means the "Bound ..." log does not fire on every rank, which will confuse anyone grepping logs
to confirm all ranks bound.

### 7. Since MPI is out of scope, do not `xfail` those tests

`tests/test_dp_modes.py:302` and seven more

Given the launcher path is not supported standalone, `xfail` is the wrong marker. It says
"expected to fail, should eventually pass". Skip or delete is the honest signal.

Three concrete things:

- The reason string is inaccurate. "`vllm_tt_plugin.launcher` imports legacy vllm core-engine
  objects" reads as an upstream deprecation. `CoreEngineLauncher` and `EngineLaunchPlan` are
  not upstream objects at all: `git log -L 82,95:vllm/v1/engine/utils.py` in the tenstorrent
  fork attributes both, plus `ParallelConfig.engine_core_launcher_cls`, to TT's own fork commit
  `ae0f0730b`. Somebody will go looking for an upstream removal that never happened. Say
  instead that the launcher needs fork-only symbols and the MPI path is unsupported standalone.
- Collapse the eight duplicated markers and eight duplicated function-local
  `from vllm_tt_plugin.launcher import parse_tt_mpi_params` lines into one module-level
  `pytest.importorskip` / `pytestmark`. Right now nothing in the file explains why those
  imports are function-local, and the first person who tidies them back to the top breaks
  collection for all 22 tests in the file, including the 14 unrelated ones.
- Two tests still assert the launcher wiring and still pass:
  `test_single_host_standard_dp_uses_upstream_launcher` (`:164`) and
  `test_rank_binding_keeps_tt_launcher` (`:188`) both assert `engine_core_launcher_cls` against
  a `SimpleNamespace`. Skipping the eight while leaving these two green is inconsistent, and
  they are asserting a qualname nothing resolves.

### 8. Test leaks `TT_VISIBLE_DEVICES` into the rest of the session

`tests/test_visible_devices.py:236`

`_bind_visible_devices_env` writes `os.environ[evar]` directly, outside monkeypatch's tracking.
For the `inherited=None` parameter the test only calls `monkeypatch.delenv(EVAR, raising=False)`,
and pytest's `MonkeyPatch.delitem` records an undo entry only when the key is present. On a box
where the var is unset, which is the normal case, nothing is registered, so the
`os.environ[EVAR] = "4,5,6,7"` the function under test performs survives teardown. Every later
test in the session then runs with `TT_VISIBLE_DEVICES=4,5,6,7`; the other parameters use
`setenv` and restore to the already-leaked value, so it persists to the end of the run.

Harmless in a host-only run today, but this is the classic passes-alone-fails-in-suite setup,
and on a 2-device box it would point later device-touching code at chips that do not exist. An
autouse fixture that snapshots and restores the var around every test in the module would cover
future tests automatically.

### 9. `tests/conftest.py` at the suite root pulls `import vllm` into the client-side suite

`tests/conftest.py:11`

`tests/tt/` is the HTTP client suite: its own conftest imports only `openai`, and the README
documents `pytest tests/tt -v --tt-server-url=... --tt-model-name=...`. With the new conftest
at `tests/` rather than next to the two modules that need it, that documented command now
imports the locally built TT vLLM during collection. Neither consumer of the `vllm_config`
fixture is under `tests/tt`.

Either make the guard lazy (import inside the fixture) or scope the conftest to the host-unit
tests.

---

## Nits

- `platform.py:121`: `if not isinstance(groups, list) or not groups: return None` collapses an
  empty list into the same `None` that means "no stored groups, keep the inherited value". A
  discovery that produced zero groups therefore takes the exact path this PR removes, silently,
  and the new `dp_index >= len(groups)` raise can never fire for `len == 0`. It is unreachable
  today only because the store site guards on truthiness, 900 lines away.
- `worker.py:74`: the signature takes `vllm_config` and `parallel_config`, where the second is
  a field of the first, and the only production call site passes both from `self`. The tests
  exploit the split to hand in an unrelated pair, so the real coupling is never exercised.
  `parallel_config` is also the only unannotated parameter in the module.
- `worker.py:101`: the `inherited == visible_devices` early return suppresses the "Bound"
  log for whichever rank already matches the launcher's export, which is typically the rank you
  most want confirmation for when debugging a mis-bind. Logging unconditionally costs nothing.
- `tests/conftest.py:56`: worth calling out in the PR description that this fixture is not a
  pure move. `structured_outputs_config` is newly added and is required by
  `platform.py:843`, which already dereferenced it unconditionally on `main`, so several tests
  routed through `register_dummy_model` were erroring before this PR, not passing.
  `world_size` and `assigned_physical_gpu_ids` are needed by the new upstream-handoff test. A
  reviewer reading the diff has no way to tell those apart from noise.
- `tests/test_visible_devices.py:7`: the module docstring says "the handoff to vLLM's
  device-index assignment", but you renamed that test away from `get_device_indices` in this
  same PR. Should say `assigned_physical_gpu_ids`.
- `tests/test_visible_devices.py` is flat functions, `tests/test_dp_modes.py` is one class, and
  `tests/test_galaxy_dp_conversion.py` is a third DP-adjacent file. The split boundary is not
  obvious either: `test_dp_modes.py:149` asserts
  `TTPlatform._standard_dp_visible_device_groups`, which is the new file's stated subject. One
  line per module docstring saying what belongs where would settle it.
- No `[tool.pytest.ini_options]` anywhere, so `xfail_strict` is off. Less important if the
  markers become skips per item 7, but worth adding regardless.

src/vllm_tt_plugin/platform.py
```diff
@@ -728,7 +731,13 @@ class TTPlatform(Platform):
...
 
     @classmethod
-    def device_id_to_physical_device_id(cls, device_id: int):
+    def device_id_to_physical_device_id(cls, device_id: int) -> str | int:
+        """Map a DP rank to its whole comma-joined TT device group.
+
+        Deviates from upstream, where ``device_id`` is a device index and the
+        return is one physical id. ``assigned_physical_gpu_ids`` therefore holds
+        a single group string, not the ``list[int]`` upstream declares.
+        """
```
viktorpusTT, line 742:
```suggestion
        """Map a local DP rank to its whole comma-joined TT device group.

        Deviates from upstream, where ``device_id`` is a logical device index and
        the return is one physical id. Sound only because TT pins ``world_size``
        to 1, so upstream asks for exactly one id per DP rank and keeps it as a
        one-element list: ``assigned_physical_gpu_ids`` carries a group string,
        not the ``list[int]`` it declares. Anything reading that list as
        per-device, its length as a device count or its entries as ints, is
        wrong for TT.
        """
```

src/vllm_tt_plugin/worker.py
```diff
@@ -71,39 +72,54 @@ class CompilationTimes(NamedTuple):
...
-    if groups is None or dp_index >= len(groups):
-        return  # no stored groups or index out of range
+    if assigned_physical_gpu_ids:
+        visible_devices = ",".join(str(d) for d in assigned_physical_gpu_ids)
+    else:
+        groups = _load_standard_dp_visible_groups(vllm_config)
+        if groups is None:
+            return
+
+        # Same index space as the mesh selection in ``init_device``.
```
viktorpusTT:
```suggestion
        # Index the fallback the way upstream indexes the primary channel:
        # device_id_to_physical_device_id(local_dp_rank) at world_size 1. Local
        # equals global because discovery only runs on a single host.
```

viktorpusTT:
Thanks for this. I have one meaningful finding and some nits:

### The reorder makes the worker's second discovery run on a narrowed cluster, and it clobbers the discovered submesh shapes

`src/vllm_tt_plugin/worker.py:218-227`, `src/vllm_tt_plugin/platform.py:1012-1025`

**Background.** `check_and_update_config` runs twice per rank. Once in the API-server
process, where `_resolve_standard_dp_visible_device_groups` spawns the discovery helper,
opens the parent mesh, calls `create_submeshes`, and stores two things on
`additional_config`: the per-rank group list (`_tt_standard_dp_visible_groups`) and a
`group string -> (rows, cols)` map (`_tt_standard_dp_mesh_grids`). Then again inside
`TTWorker.init_device`, because the worker subprocess has to re-apply the platform's config
mutations. That second call is not guarded against re-entry: the only guards inside
`_resolve_standard_dp_visible_device_groups` (`platform.py:193-196`) are
`data_parallel_size <= 1` and `_uses_explicit_tt_mpi_launch`, and in a standard-DP worker
neither holds. So each worker re-runs full discovery, including a second `open_mesh_device`.

Before this PR that was merely wasteful. `_ensure_visible_devices_env` ran *after*
`check_and_update_config`, so the worker's re-discovery saw the launcher's inherited value,
i.e. the whole machine, recomputed exactly the same groups and the same grid map, and the
overwrite at `platform.py:1019-1021` was a no-op. This PR binds first, so the re-discovery
now sees only this rank's chips and computes a different, wrong answer.

**Mechanism, Galaxy `dp=2`, 32 chips, 16 per rank.**

1. Parent: `_resolve_parent_mesh_grid("TG", 32)` gives `(4, 8)`; `create_submeshes` takes
   the `num_devices == 32` branch and uses `_galaxy_data_parallel_submesh_shape(16)`, which
   returns `MeshShape(2, 8)` (`generator.py:2986-2993, 3007-3011`). The stored map is
   `{"<16 ids>": (2, 8)}` per group. This map is the only carrier of the true submesh shape.
2. Worker: `_bind_visible_devices_env` sets `TT_VISIBLE_DEVICES` to this rank's 16 ids.
3. Worker `check_and_update_config`: `platform.py:1013` restores the correct map from
   `additional_config`, then `platform.py:1014` re-runs discovery.
4. The discovery child now sees `ttnn.get_num_devices() == 16`.
   `_resolve_parent_mesh_grid("TG", 16)` gives `(4, 8)`, whose product 32 does not match 16,
   so it falls back to `(1, 16)`. `create_submeshes` no longer takes the 32-device branch and
   returns `MeshShape(1, 16 // 2) = (1, 8)` twice, keyed by 8-chip sub-sub-groups.
5. `platform.py:1019-1021` overwrites the restored map with that, and re-stores it.
6. `_resolve_mesh_grid` (`worker.py:137-140`) looks up this rank's actual 16-id visible
   string, misses, and falls through: `_parse_mesh_grid("TG", 16, tg_mesh_grid=(8, 4))` gives
   `(8, 4)`, product 32 does not match 16, so `(1, 16)`. `MESH_DEVICE` unset gives the same
   `(1, 16)`.

The worker opens a `(1, 16)` mesh where discovery decided `(2, 8)`.

**Scope** This only diverges when a rank's submesh is 2D. T3K `dp=2` gives
4 chips per group and both paths land on `(1, 4)`; Galaxy `dp=4` gives `(1, 8)` both ways.
The bite is Galaxy `dp=2` (16 chips per group, `(2, 8)`) and any future shape that is not a
single row, which is exactly the case the grid map was added for. I have not run Galaxy
`dp=2`; this is read out of the code paths above, and the PR's own test plan still has
"DP > 1 ranks bind distinct device groups" unchecked. Separately, the redundant per-rank
mesh open is a real startup cost that this finding's fix removes for free.

**Fix.** Do not re-discover in the worker. The groups and grids were pickled in with
`VllmConfig`; consume them. Options, cheapest first: skip
`_resolve_standard_dp_visible_device_groups` when `_load_standard_dp_visible_groups` already
returns a list; or gate discovery on being the parent; or at minimum refuse to overwrite a
loaded grid map that already contains an entry for the current visible group.

### `--device-ids` silently wins over the discovered groups and binds one chip per rank

This switch actually cannot do for TT what it is meant to do for GPU.
Please just create an issue to reject this.

### `tests/test_mpi_launcher.py` can never run in this repo, so delete it instead of skipping it


Out of scope for this PR, see follow-up B: `src/vllm_tt_plugin/launcher.py` itself has the
same problem as its tests, and `check_and_update_config` still points
`engine_core_launcher_cls` at it.

### Test isolation across the newly split modules

`check_and_update_config` writes class-level state on `TTPlatform`
(`_standard_dp_visible_device_groups`, `_standard_dp_mesh_grids`, `sample_on_device_mode`,
`always_compat_sampling`), and tests assert on it after the `monkeypatch.context()` exits,
so it leaks between tests and now between files: the tests that populate it via
`check_and_update_config` are in `test_dp_modes.py`, while the tests that read it through
`_resolve_mesh_grid` moved to `test_visible_devices.py`. It happens to be safe today because
`register_dummy_model` patches discovery to return plain strings, so
`_split_standard_dp_discovery_result` yields an empty grid map and nothing is stored. That is
luck, not isolation. Since this PR is splitting the modules, add an autouse fixture in
`conftest.py` that resets those class attributes.

viktorpusTT (approved):

PR #32, opened 2026-08-10
feature: serve DiffusionGemma block outputs on vLLM 0.25.1

zhenghaoniTT:
## Summary
- register DiffusionGemma block serving in the standalone TT plugin while preserving the model-owned 256-token canvas contract
- serve DiffusionGemma with the engine-based `gemma4` reasoning/tool parsers that ship with vLLM 0.25.1 (the same ones upstream's DiffusionGemma recipe uses), validated against real block outputs on P150x4. The plugin's own `gemma4` parser copies, the interim `diffusion_gemma` aliases, and their tests are **deleted**: both DiffusionGemma and autoregressive Gemma 4 now resolve upstream's parsers with no plugin-side registration
- synchronize request admission with auto-fitted `max_model_len` (`--max-model-len -1` now fits the block-output KV budget minus one output canvas instead of failing startup), resolve omitted output limits on request clones, and reject unsupported resumable streaming-input sessions before EngineCore
- make a forced prefix-cache reset succeed for block models: the engine layer aborts running block requests and notifies the owning clients before resetting; engines without a client-notification path (bare in-process `EngineCore`) refuse instead, and `pause_generation(mode="keep")` resets return `False` without touching retained requests
- fail fast for unsupported explicit MPI launch and verify model-owned request/capture release through scheduler and worker shutdown paths

CI: device run for the current SHA (tt-metal `main`, which now includes the DiffusionGemma runtime): https://github.com/tenstorrent/tt-shield/actions/runs/32145555857 — in progress (an earlier run validated GPQA 90 vs GPU baseline 70 and the full param-conformance suite on this SHA; its only blocker was a benchmark-harness transformers pin, fixed in tt-inference-server). Host-side plugin suite (224 tests) passes on vLLM 0.25.1 + ttnn.

## Sampling controls (`temperature` etc.)

- Requests carrying non-default sampling controls (`temperature`, `top_p`, `top_k`, `min_p`, `seed`, and the presence/frequency/repetition penalties) are **not rejected**. For compatibility with common OpenAI clients they are accepted as-is, then neutralized on the per-request `SamplingParams` clone (e.g. `temperature` reset to `1.0`) after vLLM admits the request — the caller-owned object is never mutated — with a `warning_once` ("HTTP sampling controls are accepted but ignored") and the per-field details logged at debug level.
- The effective sampling temperature is entirely model-owned: DiffusionGemma's denoise loop on the tt-metal side has its own Gumbel sampler and internal temperature schedule, and per-request temperature cannot be passed through from the vLLM layer — so per-request temperature control is currently not possible.
- This only applies to block-output models; autoregressive models keep the existing per-request sampling behavior unchanged.


docs/diffusion-gemma.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Installation
+
+Install vLLM 0.24.0 into an activated tt-metal environment, then install this
+plugin. Do not install a CUDA vLLM wheel over that environment.
+
+```bash
+VLLM_TARGET_DEVICE=empty uv pip install --no-binary vllm vllm==0.24.0
+uv pip install -e .
+```
```
viktorpusTT: This is different from the main `README.md` instructions. We shouldn't duplicate this.

src/vllm_tt_plugin/platform.py
```diff
@@ -1100,6 +1324,113 @@ def validate_request(
...
+        # The model owns its Gumbel sampler and temperature schedule. Common
+        # OpenAI clients still send transport sampling controls, so accept and
+        # neutralize those values before InputProcessor clones SamplingParams.
+        # This intentionally mutates the caller-owned object; server requests
+        # construct a fresh object, while offline callers should not reuse it
+        # across block-output and autoregressive models.
+        ignored = []
+        if params.temperature != 1.0:
+            ignored.append(f"temperature={params.temperature!r}")
+            params.temperature = 1.0
```
viktorpusTT: These neutralizations modify the caller's live object, as you note in the .md.

The PR already owns the better seam. _install_block_output_input_processor_patch wraps
process_inputs and holds request.sampling_params, which is the clone. Move all eight neutralizations there. validate_request goes back to being pure and shrinks from 119 lines to something reviewable. The apology comment at 1393 and the ordering constraint it creates with the rejection list both disappear.

src/vllm_tt_plugin/logger.py
```diff
@@ -35,4 +51,9 @@ def init_tt_logger(name: str) -> logging.Logger:
     )
     if not any(isinstance(f, _TTFileTag) for f in logger.filters):
         logger.addFilter(_TTFileTag())
+    # vLLM patches warning_once/info_once per instance in init_logger; this
+    # stdlib logger needs its own so *_once call sites work on every model.
+    if not hasattr(logger, "warning_once"):
```
viktorpusTT: `debug_once` missing. vLLM patches three methods. The plugin installs two. Any TT call site copying a vLLM idiom with `debug_once` raises.

Guard each method independently, add `debug_once`, and rewrite the test to (a) `monkeypatch.setattr(logging.Logger, "warning_once", ...)` to simulate the patched-class process and (b) use `caplog` to assert exactly one record survives two identical `warning_once` calls.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -169,3 +173,64 @@ def _schedule_decode_only(self) -> SchedulerOutput:
...
+    def reset_prefix_cache(
+        self, reset_running_requests: bool = False, reset_connector: bool = False
+    ) -> bool:
+        """Avoid a stale-canvas resume that vLLM's AR reset cannot represent."""
+        if self._is_block_output_model and reset_running_requests and self.running:
+            logger.error(
+                "Cannot reset prefix cache while a block-output request is "
+                "running; finish or abort the request first."
+            )
+            return False
```
viktorpusTT: This is unfortunately ignored in the caller. We should raise, just as the standard scheduler does.

src/vllm_tt_plugin/platform.py
```diff
@@ -942,6 +1072,75 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+            # DiffusionConfig(canvas_length=256). Scheduler interprets that
+            # canvas as speculative draft tokens, while this plugin separately
+            # reserves the model-owned output block. Clear the upstream
+            # diffusion path so one physical canvas is accounted exactly once.
+            if vllm_config.diffusion_config is not None:
+                logger.info(
+                    "Block-output model owns diffusion scheduling; disabling "
+                    "upstream DiffusionConfig speculative-token accounting."
+                )
+                vllm_config.diffusion_config = None
```
viktorpusTT: `ModelConfig.is_diffusion` is untouched, resulting in 3 compensations: `VLLM_USE_V2_MODEL_RUNNER`, `TTWorker.take_draft_token_ids()` returning `None` and `
extra_placeholders = K - self.num_sampled_tokens_per_step`.
Better to make `is_diffusion` uniformly false. The hook already rewrites `hf_config.architectures` in place, so deleting or renaming `hf_config.canvas_length` in the same place is a two-line change that removes all three compensations at once.

src/vllm_tt_plugin/platform.py
```diff
@@ -429,6 +436,53 @@ def tokenizer_args_from_config_tt(config, **kwargs):
...
+
+    if hasattr(input_processor, "_tt_original_process_inputs"):
+        return
+
+    original = input_processor.InputProcessor.process_inputs
+    input_processor._tt_original_process_inputs = original
+
+    def process_inputs_tt(self, request_id, prompt, params, *args, **kwargs):
+        output_size = get_tt_output_tokens_per_step(self.vllm_config)
+        if output_size > 1 and kwargs.get("resumable", False):
```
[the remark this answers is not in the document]
zhenghaoniTT: Updated: the wrapper now intentionally checks only `kwargs.get("resumable", False)`. Signature introspection and the positional regression were removed because production callers pass this flag by keyword.

docs/diffusion-gemma.md
zhenghaoniTT: Removed the duplicated commands. The Installation section now points to the README Environment Setup and `source docs/install-vllm-tt.sh`.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Moved all eight neutralizations into the `process_inputs` wrapper so they operate on the cloned `request.sampling_params`. `validate_request` remains pure, and tests verify the caller-owned object is unchanged.

src/vllm_tt_plugin/logger.py
zhenghaoniTT: Added `debug_once` and guarded all three once methods independently. Tests cover a class-level `warning_once`, duplicate suppression through `caplog`, and isolate the shared once-cache between tests.

src/vllm_tt_plugin/async_decode.py
```diff
@@ -42,6 +43,21 @@ class TTFinalizedDecode:
     tt_log_probs: torch.Tensor | None
 
 
+def model_accepts_kwarg(fn: Any, name: str) -> bool:
```
viktorpusTT: Delete this, nothing in tt-metal needs it. If you keep it for some, make the mismatch raise when a non-identity `slot_remap` or a non-default `empty_slots` was computed, and gate on a declared  `model_capabilities` flag rather than `inspect.signature`. If any form of the gate survives, test all three shapes: strict signature, explicit named parameter, `**kwargs`.

src/vllm_tt_plugin/model_runner.py
```diff
@@ -595,6 +644,21 @@ def _build_per_layer_specs(
...
+    def _release_model_request(self, req_id: str) -> None:
+        """Release model-owned state while the slot mapping is still valid.
+
+        State follows the request's ``_req_state_slot`` slot, not its batch
+        row: prefill can park a request at a slot other than its row when the
+        preferred row is held (``_alloc_prefill_state_slots``). The row is only
+        a fallback for requests that never went through slot allocation.
+        """
+        slot = self._req_state_slot.get(req_id)
+        if slot is None:
```
viktorpusTT: This is a different index space. Delete the row fallback. Release only when `_req_state_slot` has an entry.

src/vllm_tt_plugin/platform.py
```diff
@@ -942,6 +1072,75 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
             model_class, "model_capabilities", None
         )
+        output_tokens_per_step = cls._resolve_output_tokens_per_step(model_class)
+        is_block_output_model = output_tokens_per_step > 1
+        if is_diffusion_gemma and not is_block_output_model:
+            raise ValueError(
+                "DiffusionGemma must declare output_tokens_per_step > 1 "
+                "in model_capabilities"
+            )
+        if is_block_output_model:
```
viktorpusTT, line 1608: Assert that the resolved KV budget covers `max_model_len + output_tokens_per_step`, so a future `max_num_seqs > 1` or a fallback in `get_max_tokens_all_users` fails at launch with a clear message
 instead of stalling a request in WAITING forever, which is precisely the failure DG's `get_kv_cache_spec` docstring records having hit before.

src/vllm_tt_plugin/platform.py
```diff
@@ -1100,6 +1324,113 @@ def validate_request(
         if isinstance(params, SamplingParams) and params.prompt_logprobs is not None:
             raise ValueError(f"Not yet supporting prompt_logprobs on {dev}")
 
+        output_size = cls.output_tokens_per_step
+        if not isinstance(params, SamplingParams) or output_size == 1:
+            return
+
+        prompt_token_ids = processed_inputs.get("prompt_token_ids")
```
viktorpusTT: For an embeds input, processed_inputs carries `prompt_embeds` and no `prompt_token_ids`, so the whole check is skipped. The compensating check in the input-processor patch only fires when `max_tokens` was omitted. So an embeds request with an explicit `max_tokens` reaches `EngineCore` with a canvas requirement that can exceed `max_model_len`.

Derive the prompt length the way the processor does (length_from_prompt_token_ids_or_embeds) so embeds get the same admission check. Or reject embeds inputs for block models outright, which is defensible given everything else this path rejects.

src/vllm_tt_plugin/platform.py
```diff
@@ -429,6 +436,66 @@ def tokenizer_args_from_config_tt(config, **kwargs):
...
+
+    if hasattr(input_processor, "_tt_original_process_inputs"):
+        return
+
+    original = input_processor.InputProcessor.process_inputs
+    input_processor._tt_original_process_inputs = original
+
+    def process_inputs_tt(self, request_id, prompt, params, *args, **kwargs):
+        output_size = get_tt_output_tokens_per_step(self.vllm_config)
+        resumable = kwargs.get("resumable", args[7] if len(args) > 7 else False)
```
viktorpusTT: Nothing in the code says so, and no caller exercises the args[7] option, all pass by keyword. So it is unexercised defensive code whose only possible effect is to be wrong.
Keep `kwargs.get("resumable", False)`

src/vllm_tt_plugin/entrypoints.py
```diff
@@ -29,6 +34,11 @@ def _register_tt_reasoning_parsers() -> None:
         "vllm_tt_plugin.gemma4_reasoning_parser",
```
viktorpusTT: Not needed now?

src/vllm_tt_plugin/entrypoints.py
```diff
@@ -45,6 +58,11 @@ def _register_tt_tool_parsers() -> None:
         "vllm_tt_plugin.gemma4_tool_parser",
```
viktorpusTT: Not needed now?

src/vllm_tt_plugin/platform.py
```diff
@@ -429,6 +436,66 @@ def tokenizer_args_from_config_tt(config, **kwargs):
...
+        request = original(self, request_id, prompt, params, *args, **kwargs)
+
+        cloned_params = request.sampling_params
+        if (
+            unresolved_max_tokens
+            and output_size > 1
+            and cloned_params is not None
+            and cloned_params.max_tokens is not None
+        ):
+            rounded = cloned_params.max_tokens // output_size * output_size
```
viktorpusTT: Three slightly different roundings in this file, three error messages. Unify.

tests/test_block_request_validation.py
```diff
@@ -0,0 +1,440 @@
...
+def test_short_logical_request_reserves_one_physical_canvas():
+    _validate(SamplingParams(max_tokens=16), prompt_len=768)
+
+
+def test_rounded_physical_capacity_is_enforced():
+    with pytest.raises(ValueError, match=r"requires 512 physical.*exceeding"):
+        _validate(SamplingParams(max_tokens=257), prompt_len=513)
+
+
+def test_auto_fitted_frontend_limit_is_read_live(monkeypatch):
```
viktorpusTT, line 108: This PR adds `TTPlatform.output_tokens_per_step` and `TTPlatform.block_model_config`, both written by `check_and_update_config`, and does not extend `_TT_PLATFORM_CONFIG_ATTRS`. A file-local `autouse` fixture masks it today; the next test file that calls `check_and_update_config` with a block model leaks `output_tokens_per_step=256` globally.
Add both ClassVars to `_TT_PLATFORM_CONFIG_ATTRS`; add an autouse fixture that saves and restores `InputProcessor.process_inputs` plus the sentinel; restore `VLLM_USE_V2_MODEL_RUNNER` in the same fixture.

zhenghaoniTT: Added both ClassVars to `_TT_PLATFORM_CONFIG_ATTRS`. The root autouse fixture restores `InputProcessor.process_inputs`, its sentinel, and `VLLM_USE_V2_MODEL_RUNNER` after every test.

src/vllm_tt_plugin/entrypoints.py
zhenghaoniTT: Kept this existing `gemma4` registration to preserve autoregressive Gemma 4 tool-calling behavior. DiffusionGemma now uses a separate `diffusion_gemma` alias pointing to vLLM 0.24’s official engine tool parser, so block serving does not change existing Gemma 4 behavior.

src/vllm_tt_plugin/entrypoints.py
zhenghaoniTT: Kept this existing `gemma4` registration to preserve autoregressive Gemma 4 behavior. DiffusionGemma now uses a separate `diffusion_gemma` alias pointing to vLLM 0.24’s official engine reasoning parser, so block serving does not depend on or change this legacy parser.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Block setup now deletes `hf_config.canvas_length`, invalidates cached `ModelConfig.is_diffusion`, and clears any pre-created `DiffusionConfig`. The V2 env override and draft-token RPC are removed; the scheduler adds K-1 placeholders on top of upstream’s normal one.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: All default fitting and explicit admission now go through `_resolve_block_output_max_tokens`, with one physical-canvas calculation and one error format.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Admission now derives prompt length with `length_from_prompt_token_ids_or_embeds`, so embeds and token IDs share the same physical-canvas check. Added an explicit embeds + `max_tokens` regression.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Block sizing now reserves one full output canvas per user and asserts that resolved capacity covers `max_model_len + output_tokens_per_step`. Tests cover both a valid model budget and an undersized fallback.

src/vllm_tt_plugin/model_runner.py
zhenghaoniTT: Removed the batch-row fallback. `release_request` is called only when `_req_state_slot` contains the request; a regression covers the unmapped-row case.

src/vllm_tt_plugin/async_decode.py
zhenghaoniTT: Removed `model_accepts_kwarg` and all signature inspection. Computed `slot_remap` and `empty_slots` are always forwarded, with regressions for both paths.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: The wrapper now uses only `kwargs.get("resumable", False)`. Signature introspection and the positional regression were removed.

src/vllm_tt_plugin/scheduler.py
zhenghaoniTT: Block-output resets with running requests now raise `RuntimeError`; the regression test verifies the request remains running.

zhenghaoniTT:
https://github.com/tenstorrent/tt-shield/actions/runs/31919097339
new ci passed

tests/test_lane_model_runner.py
viktorpusTT, line 256: Add `_is_block_output_model=False` and `_output_tokens_per_step=1` to the stub runner in this test. Otherwise it gives `AttributeError: 'types.SimpleNamespace' object has no attribute '_is_block_output_model'`

src/vllm_tt_plugin/scheduler.py
```diff
@@ -204,3 +208,61 @@ def _schedule_decode_only(self) -> SchedulerOutput:
             if partial_prefills:
                 self.running.extend(partial_prefills)
         return result
+
+    def reset_prefix_cache(
+        self, reset_running_requests: bool = False, reset_connector: bool = False
+    ) -> bool:
+        """Avoid a stale-canvas resume that vLLM's AR reset cannot represent."""
+        if self._is_block_output_model and reset_running_requests and self.running:
```
viktorpusTT, line 459: My previous comment was not correct. The reset should be made to succeed: abort or drain the running block requests before delegating to `super()`. If that is too large for this PR, restrict the raise to the synchronous utility RPC (which `_invoke_utility_method` catches at `core.py:1448`) and return `False` on the deferred callback path, restoring the old behavior there.

src/vllm_tt_plugin/platform.py
```diff
@@ -966,6 +1140,88 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
+
+            # Upstream 0.24 detects diffusion from ``hf_config.canvas_length``.
+            # TT block models own their canvas lifecycle, so remove that marker
+            # and invalidate the cached property before vLLM selects a model
+            # runner or constructs the scheduler. This keeps every downstream
+            # view of ModelConfig.is_diffusion uniformly false.
+            hf_config = model_config.hf_config
+            if hasattr(hf_config, "canvas_length"):
+                delattr(hf_config, "canvas_length")
+                model_config.__dict__.pop("is_diffusion", None)
```
viktorpusTT:
```
declared = hf_config.__dict__.pop("canvas_length", None)   # cannot raise; covers class attrs and aliases
if declared is not None and declared != output_tokens_per_step:
    raise ValueError(...)                                   # validated handoff, not a silent overwrite
model_config.__dict__.pop("is_diffusion", None)             # unconditional, outside the guard
assert not model_config.is_diffusion
```

src/vllm_tt_plugin/scheduler.py
```diff
@@ -204,3 +208,61 @@ def _schedule_decode_only(self) -> SchedulerOutput:
...
+        """Reserve the complete physical output emitted by each block step.
+
+        The platform removes the upstream diffusion marker, so vLLM reserves
+        its normal one sampled-token placeholder. The TT adapter returns one
+        K-token canvas; reserve the remaining K-1 positions.
+        """
+        super()._update_after_schedule(scheduler_output)
+        if not self._is_block_output_model:
+            return
+        extra_placeholders = self._output_tokens_per_step - 1
```
viktorpusTT: Restore `extra_placeholders = self._output_tokens_per_step - self.num_sampled_tokens_per_step` (identical value today, correct by construction) and add
`assert self.num_sampled_tokens_per_step == 1` in `__init__` for block models so the cross-module contract fails at construction. Keep the docstring.

src/vllm_tt_plugin/model_runner.py
viktorpusTT, line 1759: _build_chunked_prefill_output duplicates the canonical
builder but drops the width normalization: :1754 does
sampled_token_ids.view(num_reqs) and :1758 emits [int(...)], both of which
assume one token per request. The same divergence exists at :2043, where the
all-intermediate shortcut produces torch.zeros(sz) while the block path at
:2014 produces (0, K).

The two features are disjoint only because a config gate in a different file
(platform.py:1154) disables chunked prefill for block models. Nothing in
model_runner.py records that. A block model reaching this path would silently
truncate its 256-token canvas to one token rather than fail.

Fix. Fold the intermediate-row suppression into _build_runner_output as a
mask parameter, so there is one builder that honors _output_tokens_per_step.
If that is too invasive here, add assert self._output_tokens_per_step == 1 on
entry to _build_chunked_prefill_output so the coupling is local and loud.

tests/test_block_scheduler.py
```diff
@@ -0,0 +1,293 @@
...
+
+BLOCK_SIZE = 16
+CANVAS = 16
+MAX_MODEL_LEN = 256
+
+
+def _scheduler(
+    output_width: int = CANVAS, *, diffusion_checkpoint: bool = False
+) -> TTScheduler:
+    model_config = ModelConfig(
```
viktorpusTT, line 80: This is offline-hostile and rate-limit hostile.

Point model= at a small vendored config under tests/, or supply a local PretrainedConfig through a module-scoped fixture. Keep the object a real ModelConfig rather than a stub: that is what makes the is_diffusion chain honest, and it is worth keeping.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -204,3 +208,61 @@ def _schedule_decode_only(self) -> SchedulerOutput:
...
+        if len(new_token_ids) != self._output_tokens_per_step:
+            raise ValueError(
+                "Model output width violates output_tokens_per_step: "
+                f"{len(new_token_ids)} != {self._output_tokens_per_step}"
+            )
+
+        # Scheduler appends token-by-token and trims at EOS, stop tokens,
+        # max_tokens, or max_model_len. The reservation is physical, so consume
+        # all K placeholders even when the client-visible block is trimmed.
+        new_token_ids, stopped = Scheduler._update_request_with_output(
```
viktorpusTT, line 516: `AsyncScheduler._update_request_with_output` also calls `self.kv_cache_manager.cache_blocks(...)`. The bypass to grandparent here silently skips it. That is harmless only because `cache_blocks` early-returns when caching is off.

Keep the grandparent call. Add a comment naming `cache_blocks` and citing `platform.py:1194` as the reason skipping it is safe, add `assert not self.kv_cache_manager.enable_caching` for block models in
`__init__`, and add `assert cache_calls == []` to the K>1 test so the skip reads as intent.

src/vllm_tt_plugin/entrypoints.py
viktorpusTT: gemma4 should resolve to upstream's parser, and the plugin should stop shipping a competing implementation of a protocol vLLM now owns. Concretely:

Delete both gemma4 `register_lazy_module` calls in `entrypoints.py`. Upstream's registration already ran by then, so gemma4 simply resolves upstream with no further code.

Delete the `diffusion_gemma` aliases too. Their only purpose is to reach the upstream parser around the plugin's override; once the override is gone they are a synonym. 

Update `docs/diffusion-gemma.md to pass `--reasoning-parser gemma4` and `--tool-call-parser gemma4`. Note this changes the documented launch command, so it needs a CI run to confirm.

Delete `gemma4_reasoning_parser.py` and `gemma4_tool_parser.py`, and with them `tests/test_gemma4_tool_parser.py`. That also removes the `Gemma4ToolParser.__init__` signature fix this PR had to make (`gemma4_tool_parser.py:46-48`, accepting DelegatingParser's positional tools), which is a good illustration of the ongoing cost: maintaining a parallel parser means tracking upstream's constructor contract by hand.

De-risk it cheaply before deleting: point the 10 existing tests in `tests/test_gemma4_tool_parser.py` at `Gemma4EngineToolParser` instead of `Gemma4ToolParser`. If they pass, the behavioral argument for keeping the plugin copy is empty and the deletion is safe.

src/vllm_tt_plugin/platform.py
```diff
@@ -966,6 +1140,88 @@ def check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
...
             model_class, "model_capabilities", None
         )
+        output_tokens_per_step = cls._resolve_output_tokens_per_step(model_class)
+        is_block_output_model = output_tokens_per_step > 1
+        if is_diffusion_gemma and not is_block_output_model:
+            raise ValueError(
+                "DiffusionGemma must declare output_tokens_per_step > 1 "
+                "in model_capabilities"
+            )
+        if is_block_output_model:
```
viktorpusTT, line 1608: `release_request` and `release_persistent_capture` are now implied by declaring `output_tokens_per_step > 1`. Validate here that a block-output model class exposes both hooks.

viktorpusTT:
`SCHEDULING.md` is now only half-true about token placeholders. Add a short "Block-output reservation" subsection stating that placeholders have two independent users: async decode lookahead, and synchronous whole-canvas reservation when output_tokens_per_step > 1. Cross-link `docs/diffusion-gemma.md`.

docs/diffusion-gemma.md
```diff
@@ -0,0 +1,125 @@
+# DiffusionGemma Block Serving
+
+DiffusionGemma is a block-output model. One model invocation commits a complete
+256-token canvas, not one autoregressive token. The paired implementation is
+`tenstorrent/tt-metal` commit
+`5fec49388893120148e189ef2602de429f214096`. Its adapter declares:
```
viktorpusTT: This commit will not make it to main after the branch is squashed. Assuming the metal branch is merged before vllm, we don't need to state the revision here. It is generally assumed that we can't use tt-metal older than vllm plugin.

README.md
viktorpusTT, line 406: We can add DiffusionGemma to this list now.

tests/test_lane_model_runner.py
zhenghaoniTT: Done — both stub runners set `_is_block_output_model=False` and `_output_tokens_per_step=1` now.

src/vllm_tt_plugin/scheduler.py
zhenghaoniTT: Made the reset succeed: the engine layer now aborts the running block requests and sends the abort outputs to their clients before delegating, so the RPC path returns True. The keep-mode pause still returns False without touching retained requests. One wrinkle: a bare in-process EngineCore has no `_send_abort_outputs`, and `finish_requests` alone emits nothing — aborting there would leave the caller waiting forever — so that path keeps refusing with the error. The scheduler guard stays as a backstop, and all paths have tests.

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Done as written, plus a regression for the declared-vs-capability mismatch.

src/vllm_tt_plugin/scheduler.py
zhenghaoniTT: Restored the expression and added the `__init__` assert; docstring kept.

src/vllm_tt_plugin/model_runner.py
zhenghaoniTT: Added the entry assert, plus one in the all-intermediate shortcut, so the coupling fails loudly. The mask-parameter refactor felt too invasive here — happy to do it as a follow-up.

tests/test_block_scheduler.py
zhenghaoniTT: Vendored a minimal qwen2 config under `tests/model_configs/` and kept the real `ModelConfig`. The file passes with `HF_HUB_OFFLINE=1` — no hub traffic left.

src/vllm_tt_plugin/scheduler.py
zhenghaoniTT: Kept the grandparent call with a comment naming `cache_blocks` and the platform gate, added the `__init__` assert for block models, and the K>1 test now asserts `cache_calls == []` (the K=1 test asserts the inverse, so the skip reads as intent).

src/vllm_tt_plugin/entrypoints.py
zhenghaoniTT: Deleted it all: both registrations, the `diffusion_gemma` aliases, the two parser modules and their tests. The docs now pass `--reasoning-parser gemma4` / `--tool-call-parser gemma4`. On the de-risk step — instead of pointing the old tests at `Gemma4EngineToolParser`, I'm leaning on upstream's own suite for that class (0.24.0 ships the gemma4 tool-parser and engine streaming tests), plus a registry-identity test pinning `gemma4` to the engine classes. Device CI ran the documented launch on this branch and the conformance suite + GPQA passed: https://github.com/tenstorrent/tt-shield/actions/runs/32095447390 (its only blocker was a benchmark-harness dep on the inference-server side, fixed there; re-run in flight).

src/vllm_tt_plugin/platform.py
zhenghaoniTT: Done

docs/diffusion-gemma.md
zhenghaoniTT: Dropped. With the metal branch merged, the doc just says "the paired tt-metal checkout".

README.md
zhenghaoniTT: Added.

src/vllm_tt_plugin/entrypoints.py
zhenghaoniTT: Superseded — this registration is deleted now; `gemma4` resolves upstream.

zhenghaoniTT:
Re: SCHEDULING.md placeholders — added a "Block-output reservation" subsection that names both users of the accounting and cross-links docs/diffusion-gemma.md, phrased as a reference to the existing async-scheduler section rather than restating it.

src/vllm_tt_plugin/entrypoints.py
zhenghaoniTT: Ran the de-risk step after all: pointed the old behavioral tests at `Gemma4EngineToolParser` (mock tokenizer mirroring upstream's own fixture) on the real 0.24 wheel. 5/9 pass verbatim. 3 more pass once the parser is constructed with the tool schemas: without one the engine leaves bare scalars as strings (`n:3` -> `"3"`) where the plugin copy guessed types, but real requests always carry the schemas, so that's parity in practice. The one genuine divergence is the unterminated-block case: the plugin parser dropped `<|tool_call>call:a{x:1` (no closing marker), while the engine salvages it as a completed call with `tools_called=True`. That only surfaces when generation truncates mid-call, and it's upstream's deliberate semantics. So nothing the plugin copy uniquely provided — the deletion stands.

zhenghaoniTT:
https://github.com/tenstorrent/tt-shield/actions/runs/32145555857
passed

viktorpusTT:
Thank you for your updates. Some more findings:

**all 16 tests in `test_block_scheduler.py` error in setup** with
```
pydantic_core._pydantic_core.ValidationError:
  Model architectures ['TTQwen2ForCausalLM'] failed to be inspected
  ... registry.py:963 ModuleNotFoundError: No module named 'models'
```

Cause: `_scheduler()` builds a real `VllmConfig`, whose `__post_init__` runs `TTPlatform.check_and_update_config`, which rewrites `architectures` to `TTQwen2ForCausalLM` (`platform.py:1188`) and then resolves that class (`platform.py:1243`), importing the tt-metal `models` package. This is the only file in `tests/` that depends on a device software stack; every other file runs against upstream vLLM with a trivial `ttnn` stand-in.

If tt-metal is present, the fixture fails later in the
same function, at its last statement. The round-3 fix for the HuggingFace Hub
dependency vendored `tests/model_configs/qwen2/` containing exactly one file,
`config.json`. `ModelConfig` (built at `tests/test_block_scheduler.py:43-47`)
leaves `skip_tokenizer_init` at its default `False` and sets `self.tokenizer =
self.model` (`vllm/config/model.py:487`), so the tokenizer path is that same
directory. Sixty lines later, `_scheduler()`'s `return TTScheduler(...)` passes
`structured_output_manager=StructuredOutputManager(config)`
(`tests/test_block_scheduler.py:104`), whose `__init__` calls
`cached_tokenizer_from_config` unconditionally when `skip_tokenizer_init` is
False (`vllm/v1/structured_output/__init__.py:70-79`), reaching a bare
`tokenizer_cls_.from_pretrained(...)` with no fallback
(`vllm/tokenizers/registry.py:228`). `AutoTokenizer.from_pretrained` on a local
directory with no `tokenizer_config.json` or `tokenizer.json` raises `OSError`.

1. In `_scheduler()`, stub model resolution the way the sibling file already
   does. `tests/test_block_request_validation.py:261-276` monkeypatches
   `get_model_architecture` and `register_tt_models`. That keeps the real
   `ModelConfig`, which is what makes the `is_diffusion` chain honest, while
   removing the tt-metal dependency.
2. Pass `skip_tokenizer_init=True` to the `ModelConfig(...)` at `:44`. Nothing in
   the file needs a tokenizer: `_request()` supplies `eos_token_id=2` explicitly
   at `:114`. Alternatively vendor a minimal tokenizer.
3. Add a `pytest tests -q --ignore=tests/tt` job that installs upstream vLLM
   (`VLLM_TARGET_DEVICE=empty`) plus a `ttnn` shim, or add the host suite to the
   tt-metal workflow's container step where vLLM and ttnn already exist. Without
   this, every finding below regresses silently, and this is the third round in a
   row where the block tests were broken at head because nobody ran them.

**The engine patches identify a block model by reading a private scheduler attribute, with a silent default, and one of the plugin's two schedulers does not have it**

In `src/vllm_tt_plugin/platform.py:632` and `:682`: use `get_tt_output_tokens_per_step(self.vllm_config) > 1`.
That is correct for `TTScheduler` and `TTLaneCoordinator` alike, removes the
private-attribute reach, and fails loudly if the config key disappears. If a
scheduler-side predicate is genuinely wanted, promote it to a public property on
both scheduler classes.

**A KV-sizing query silently rewrites `max_model_len`, and two files credit vLLM for the change**

`src/vllm_tt_plugin/worker.py:604-633`, write at `:624`; docstring at `:507-513`

Keep the sizing function pure. Move the fit into a named
`_fit_block_output_max_model_len(vllm_config, num_tt_blocks)` called explicitly
from `TTWorker.determine_available_memory`, where the worker is legitimately
allowed to adjust its own view of the config, and say in its docstring that it
deliberately pre-empts `vllm/v1/core/kv_cache_utils.py:2092` because the TT pool
is fixed rather than profiled. The `raise` half of the branch is a validation and
can stay. Update the README `-1` row the way the adjacent row was updated.

**The documented launch recipe is not the configuration that was validated, and the difference silently returns empty completions**

`docs/diffusion-gemma.md` prints an env block plus a `vllm serve` command and says
"the values above are the QB2 (`MESH_DEVICE=P150x4`) release configuration
validated by tt-shield". The validated tt-shield run sets 17 environment variables. The
doc names five. Among the omitted ones, three gate prompt admission inside the
adapter and are **off by default** on tt-metal `main`:

| Variable | Default | Source |
|---|---|---|
| `DG_UPFRONT_LAZY_PREFILL_RECAPTURE` | `"0"` | `generator_vllm.py:217` |
| `DG_UPFRONT_COARSE_PREFILL_BUCKETS` | `"0"` | `generator_vllm.py:227` |
| `DG_PREFILL_FIXED_CHUNKS` | `"0"` | `generate.py:322` |

With them at their defaults and `DG_UPFRONT_PREFILL_WARMUP_LENS=32,64,96` as the doc prescribes,
`_prefill_execution_len_is_warmed` (`generator_vllm.py:357-374`) accepts only an
execution length in the warmed set. Any other prompt hits the reject branch at
`generator_vllm.py:1440-1456`, which ends the request with a full canvas of stop
ids. The adapter's own message says it: "Ending this request with an empty
answer; the server stays up. Enable DG_UPFRONT_LAZY_PREFILL_RECAPTURE or add
{cache_len} to DG_UPFRONT_PREFILL_WARMUP_LENS."

`docs/diffusion-gemma.md:29` tells to export `VLLM_RPC_TIMEOUT` which does not exist in vLLM 0.24.

**Two sites decide whether prefix caching may stay on, they read different inputs, and the stricter one runs first**

If --no-enable-prefix-caching is not given, the adapter's declaration is not consulted at all and the startup fails. Correct approach:

Make model_capabilities["supports_prefix_caching"] the single input, and resolve
it once, before the block-output branch:

Run the capability-driven auto-disable early, so a declaration of False switches cache_config.enable_prefix_caching off for every model, block or not. --no-enable-prefix-caching then stops being a hard requirement for DiffusionGemma, and docs/diffusion-gemma.md can drop it from the launch command.

Keep exactly one raise, and point it at the contradiction rather than at the operator's setting: if output_tokens_per_step > 1 and the model declares supports_prefix_caching=True, that is a model-declaration bug, and refusing to start is the right response. Say so in the message.

Add the two tests: a block-output model declaring False starts with cache_config.enable_prefix_caching False and no exception, and a block-output model declaring True raises. Together they pin both branches and make the cleanup in step 1 safe to perform.

**One integer has two independent stores, seven names, and six spellings of the predicate derived from it**

Two adjacent lines write the same number to two independent
places:

```python
store_tt_output_tokens_per_step(vllm_config, output_tokens_per_step)
cls.output_tokens_per_step = output_tokens_per_step
```

The same integer is additionally spelled `_output_tokens_per_step`, `output_size`,
`width`, `num_out_tokens`, `output_width`, `CANVAS`, and "physical block width"
in `docs/SCHEDULING.md`. Derived from it, "is this a block-output model" is
recomputed six ways: `is_block_output_model`, `_is_block_output_model`,
`output_size > 1` / `== 1`, `output_tokens_per_step > 1`, and the architecture-name
predicate `is_diffusion_gemma`. `config.py:159-163` already houses the right shape
of helper for this (`uses_tt_lane_coordinator(vllm_config)`), so the canonical
pattern exists and was not used.

Add `is_tt_block_output_model(vllm_config) -> bool` next to
`uses_tt_lane_coordinator` and route every `> 1` comparison through it. Delete the
`TTPlatform.output_tokens_per_step` ClassVar and have the two frontend readers
take the value from the one config store (they already reach for
`cls.block_model_config`, so a config handle is at hand). Keep one validator, and
make it the reachable one. Settle on one noun in code and one in docs.

zhenghaoniTT:
Thanks for the thorough round and all six findings are addressed. 

**1. `test_block_scheduler.py` setup errors**

- `_scheduler()` now stubs model resolution the way the sibling file does: `register_tt_models` and `get_model_architecture` are patched (`tests/test_block_scheduler.py:57`, `:67`), keeping the real `ModelConfig` — and the `is_diffusion` chain honest — while removing the tt-metal dependency (`5ce8be7`).
- The `ModelConfig(...)` passes `skip_tokenizer_init=True` (`:81`); nothing in the file needs a tokenizer (`5ce8be7`).
- CI now runs the host suite: the `unit-tests` job in `.github/workflows/ci.yaml` installs upstream vLLM with `VLLM_TARGET_DEVICE=empty` through the same `docs/install-vllm-tt.sh` operators use, supplies the `ttnn` stand-in from `ci/host-stubs`, and runs `pytest tests/ --ignore=tests/tt` on Python 3.10 and 3.12 (`aa66440`). The job has already paid for itself: it caught the one semantic conflict in the 0.25.1 merge (a fixture missing the block-output width) at head, fixed in `dd0649f`.

**2. Private scheduler attribute in the engine patches**

Both patch sites derive the predicate from the config store: `is_tt_block_output_model(scheduler.vllm_config)`, preceded by `require_tt_output_tokens_per_step(...)` so a missing key fails loudly instead of defaulting (`platform.py:750-751`, `:808-809`; the same helper serves the other sites at `:657`, `:710`). No scheduler attribute is read, so it holds for `TTScheduler` and `TTLaneCoordinator` alike (`9bfa6b6`). The scheduler-side public property alternative wasn't needed.

**3. KV-sizing query rewriting `max_model_len`**

The sizing function is pure again. The fit moved into a named `_fit_block_output_max_model_len(vllm_config, num_tt_blocks)` (`worker.py:639`) whose docstring states it deliberately pre-empts vLLM's `_auto_fit_max_model_len` because the TT pool is fixed rather than profiled; the `raise` half stayed behind as validation, and the README `-1` row was updated like the adjacent row (`5ce8be7`).

One deviation from the letter of the suggestion: the call site is `TTWorker.init_device` (`worker.py:261`), not `determine_available_memory`. Upstream runs the whole executor init — `init_device`, then `load_model` — before `EngineCore.__init__` asks for available memory, so a model that sizes its own KV state at load time would otherwise see the unfitted length (for `--max-model-len -1`, the full HF context). The comment at the call site records that ordering constraint. Happy to move it if you see a way around the load-time ordering.

**4. Launch recipe vs. validated configuration**

The env block now exports the three admission gates — `DG_UPFRONT_COARSE_PREFILL_BUCKETS=1`, `DG_UPFRONT_LAZY_PREFILL_RECAPTURE=1`, `DG_PREFILL_FIXED_CHUNKS=1` — plus `DG_PREFILL_RAGGED_CHUNK` and the warmup lens, with a paragraph explaining the gates are off by default upstream and required to serve arbitrary prompt lengths. The doc also states explicitly which validated variables are omitted and why (values match the tt-metal defaults; `VLLM_ENABLE_V1_MULTIPROCESSING` is inert) (`5ce8be7`). `VLLM_RPC_TIMEOUT` is gone. A follow-up documents that coarse prefill buckets require a power-of-two `--max-model-len` (`9c8d384`).

**5. Two prefix-caching decision sites**

`model_capabilities["supports_prefix_caching"]` is the single input, resolved once before the block-output branch (`platform.py:1422`). A declaration of `False` now switches `cache_config.enable_prefix_caching` off for every model, block or not — `--no-enable-prefix-caching` stopped being a requirement and was dropped from the doc's launch command. Exactly one raise remains and it points at the contradiction: a block-output model declaring `supports_prefix_caching=True` is a model-declaration bug and startup refuses, saying so (`platform.py:1427`). Both requested tests exist and pin the branches: `test_startup_auto_disables_prefix_caching_for_block_models` and `test_startup_rejects_block_model_declaring_prefix_caching` (`tests/test_block_request_validation.py:587`, `:597`) (`5ce8be7`).

**6. One integer, seven names**

`is_tt_block_output_model(vllm_config)` lives in `config.py` next to `uses_tt_lane_coordinator`, and every `> 1` comparison routes through it. The `TTPlatform.output_tokens_per_step` ClassVar is deleted; the config store is the only store and the frontend readers take the value from it. One validator remains — `_resolve_output_tokens_per_step`, the reachable one. Code and `docs/SCHEDULING.md` settled on `output_tokens_per_step`; the `CANVAS`/`output_width` spellings are gone (`9bfa6b6`).

---

viktorpusTT:
Hi, we're getting close! Here are my findings:

**Upstream's own DiffusionGemma config hook runs before the plugin's, and only part of it is reversed**
he mechanism is "let upstream initialize, then undo the parts we know about", and the undo list is incomplete. Prevent the hook from running at all rather than reversing it: if the TT-prefixed architecture were installed before try_verify_and_update_config` (the plugin already runs code in `pre_register_and_update`), `MODELS_CONFIG_MAP` would miss and the problem disappears. Separately, either drop the bare `DiffusionGemmaForBlockDiffusion` registration with a comment saying upstream owns it, or force-register the TT target for it, but do not leave a line whose comment claims an effect it cannot have.

**All three output-width checks are tested only on the narrow side, so a wider canvas is accepted**
`src/vllm_tt_plugin/scheduler.py:481`, `src/vllm_tt_plugin/model_runner.py:135`, `:2221`
Add the mirror case at each site.

I certainly recommend you to get one or more code reviews by AI agent(s), as they can spot a number of smaller issues that I'm not listing here.

zhenghaoniTT:
The `docs/diffusion-gemma.md:110` suggestion is a false positive. In vLLM 0.25.1, OpenAI requests first pass through `api_utils.get_max_tokens()`, which caps both explicit and omitted limits with `current_platform.get_max_output_tokens()` before `SamplingParams` reaches TT validation. Only direct/offline `SamplingParams` callers bypass that frontend cap and can be rejected, so the existing documentation is correct.

zhenghaoniTT:
https://github.com/tenstorrent/tt-shield/actions/runs/32653177109

viktorpusTT:
**The decode path is never executed.** Every canvas after the first is produced by re-running the prompt. The runner decides "is this step prompt work or generation work" by asking: does this
request still have tokens whose KV state has not been computed?

```python
def _is_still_prefilling(req_id: str) -> bool:
    row = input_batch.req_id_to_index[req_id]
    return input_batch.num_computed_tokens_cpu[row] < input_batch.num_tokens[row]
```

For a block model the counters differ by 256 on every step after the first, and every step is dispatched as a prompt step. The hardware run linked on this PR proves it. Telemetry counts across the entire run (`vllm_commit: ba8da1a`, engine `v0.25.1`):

| Event | Count |
|---|---|
| `prefill_block0` | 522 |
| `decode_block` | **0** |
| `session_create` | 523 |

And three consecutive steps of a single request:

```
{"block_idx": 0, "cache_len": 192, "committed_tokens": 256, "block_latency_s": 12.23, ...}
{"block_idx": 0, "cache_len": 448, "committed_tokens": 256, "block_latency_s":  6.09, ...}
{"block_idx": 0, "cache_len": 704, "committed_tokens": 256, "block_latency_s": 11.23, ...}
```

`block_idx` never leaves 0. `cache_len` grows by exactly one canvas per step. That is
the re-prefill signature, on real hardware, at this exact commit.

This results in quadratic prefill work. Every canvas re-encodes all previous ones. At canvas  *n* the model processes `prompt + 256*(n-1)` tokens instead of 256.

This is related to PR #67, but rebasing onto `main` fixes the autoregressive case and leaves this one unchanged.

Fix: Make the predicate width-aware, since a steady-state block step legitimately has exactly one canvas
outstanding.

```python
return (
    input_batch.num_computed_tokens_cpu[row]
    + self._output_tokens_per_step
    < input_batch.num_tokens[row]
)
```

**The safety net for upstream's DiffusionGemma config hook undoes one of its six effects, and picks its branch by guessing**

The install site is already correct. Delete the duplicate install. `src/vllm_tt_plugin/platform.py:1222`, inside `pre_register_and_update`, delete `_install_diffusion_gemma_architecture_patch()`. In every real launch path `EngineArgs.__post_init__` has already loaded general plugins before `pre_register_and_update` runs, so this call is
a no-op. Also fix the stale comment at `platform.py:1212-1215`, which describes test-model registration and no longer mentions what the lines below it do.
Replace the fallback with a single raise. `src/vllm_tt_plugin/platform.py:1535-1556`. Delete the whole block, including the `resolved_arch` lookup, the `startswith("TT")` branch, the warning, and the `vllm_config.diffusion_config = None` assignment, and put this in its place:

```python
            if vllm_config.diffusion_config is not None:
                raise ValueError(
                    "Block-output models own canvas scheduling and do not "
                    "support --diffusion-config; remove the explicit "
                    "diffusion configuration"
                )
```

Test changes may be needed.

zhenghaoniTT:
Both fixed

The predicate change is in, with one addition on top of your suggestion: the pure width-aware form turned out to misroute an AR chunked-prefill tail of exactly one prompt token — it stopped counting as prefill, the mixed-batch filter then dropped the scheduled row, and the engine either died on the output row-count assert or (with `max_tokens=1`) the request hung forever waiting for a token that was never rescheduled. So the predicate keeps anything with uncomputed prompt tokens on the prefill path unconditionally, and only applies the width once the prompt is done:

```python
num_computed = input_batch.num_computed_tokens_cpu[row]
return (
    num_computed < input_batch.num_prompt_tokens[row]
    or num_computed + self._output_tokens_per_step
    < input_batch.num_tokens[row]
)
```

After merging main this subsumes the `+1` form from #67, and the mixed-batch filter is gone entirely (main deleted it, which also removes the drop-a-scheduled-row failure mode for good). Tests now pin the steady block step going to decode, a multi-canvas replay staying prefill, the one-token prompt tail staying prefill, and a two-request batch never losing a scheduled row.

viktorpusTT (approved):
Approve if the latest CI is green. One more rebase will be needed, it seems.

zhenghaoniTT:
image
image
Looks good compare with scheduled run

viktorpusTT:
Checked the run with https://github.com/tenstorrent/tt-shield/actions/runs/32904352179/job/98034367558

1. decode_block non-zero ≈ canvases − requests, there are 410 decode_block, 585 canvases, 175 requests. 585 − 175 = 410
2. block_idx advances past 0 — 100/175 requests reach block_idx ≥ 1, max 23. Per-request sequences are contiguous 0..N with no gaps, and every block's start_pos chains to the previous next_pos (0 broken links, step = 256 = one canvas).
3. prefill_block0 cache_len no longer grows by a canvas per step, cache_len equals start_pos and sits 0–31 above prompt_len (mean 11, just alignment padding). Zero +256 deltas across the 175 prefills, and exactly one prefill_block0 per request (175 prefills / 175 releases). cache_len at request_release equals the prefill value for all 175, so KV isn't growing per committed block either.
4. session_create ≈ request count ,176 vs 175 requests (vs 585 canvases). The extra one is the upfront trace-capture warmup session (prefill_execution_len: null, followed by the single upfront_capture event).

PR #34, opened 2026-08-11
Enable token-chunked prefill for Gemma 4 (port of tenstorrent/vllm#448)

viktorpusTT:
## Purpose

Port [tenstorrent/vllm#448](https://github.com/tenstorrent/vllm/pull/448) to the standalone plugin. Enables token-chunked prefill for Gemma 4, so a long prompt is split across prefill steps instead of requiring `max_num_batched_tokens >= max_model_len`.

Not a 1:1 apply: gathered DP is gone from this repo, and upstream vLLM `0.24.0` already provides a mechanism the original PR had to build itself. See **Deviations** below.

## Modifications

**Config policy (`platform.py`)**
- `_apply_chunked_prefill_policy` replaces the unconditional force-disable. `gemma4` / `gemma4_unified` keep chunked prefill and their requested token budget; every other model type gets it turned off with the existing `max_num_batched_tokens` bump to `max_model_len`.
- Zero `long_prefill_token_threshold` for those other model types. The base scheduler applies that cap *before* it consults `enable_chunked_prefill`, so leaving it set still splits a prefill for a model whose generator cannot resume one.
- Set `disable_chunked_mm_input = True` so a chunk boundary never lands inside a multimodal item.

**Scheduling (`scheduler.py`, `lane_scheduler.py`)**
- A running `is_prefill_chunk` request is prefill work: `_has_pending_prefill` counts it, `_schedule_prefill_only` keeps it visible (hiding only pure decodes, and shrinking `max_num_running_reqs` by the hidden count), `_schedule_decode_only` hides it.
- Decode fallbacks now test for *running decodes*, not just a non-empty `running`. A partial prefill cannot be advanced by a decode step, so counting it would burn a step or livelock. Same fix in `TTLaneCoordinator.schedule` and `_local_prefill_intent`, where a continuation forces prefill intent even at lane capacity.

**Execution (`model_runner.py`, `input_batch.py`, `model_input.py`)**
- `is_prompt` detection recognises continuations (cached, not resumed, `num_computed < num_prompt_tokens`), and the mixed-batch decode filter no longer treats them as strays.
- `prompt_lens` is the *end position of the scheduled chunk* (`num_computed + num_scheduled`), not the sequence length, so the generator's `tokens[start_pos:prompt_lens]` slice processes exactly this chunk. Unchunked prefill, APC hits, and preemption replays are unchanged by this.
- New `TTModelInput.intermediate_prefill_mask` marks rows whose chunk does not finish the prompt. Those rows force host sampling (device sampling would advance device RNG state for them), get a generator clone via `_build_host_generators` / `build_merged_sampling_metadata(non_sampling_rows=...)` so a seeded request's stream does not drift, and report `[]` instead of a token via `_build_chunked_prefill_output`.
- Dropped the `assert enable_chunked_prefill is False` guard in the runner init.

**Docs**: `README.md` and `docs/SCHEDULING.md` updated; both previously stated chunked prefill was unsupported.

## Deviations from tenstorrent/vllm#448

1. **Gathered DP dropped.** `engine.py`'s `DPGatherHandle` / `dp_gather_*` and the `worker.py` / `model_runner.py` gather hooks are unreachable in this repo (`platform.py` points `dp_engine_core_proc_cls` at upstream's `DPEngineCoreProc`; nothing imports `vllm_tt_plugin.engine`). The PR's changes to those paths, and its `tests/test_dp_engine.py`, are not ported. Standard multi-process DP runs independent per-rank engines, so each rank handles its own chunking through the non-DP path.
2. **Preemption fix dropped entirely.** #448 carried a `_PendingOutputs` class that discarded a preempted request's in-flight async decode tokens. Not ported: on vLLM `0.24.0` those tokens are valid at preempt time (the forward completed before any block was freed, device submits form a strict queue, and every async op is forced to complete before the next prefill), the base class appends them and the resumed prefill replays them, so discarding them truncates the response by one token per preemption. `Request.async_tokens_to_discard` is written in exactly one place in `0.24.0`, inside `reset_prefix_cache`, and is not a general stale-output mechanism. The invariant is recorded in the `TTScheduler` docstring.
3. **Merged with this repo's local scheduler work.** `_has_pending_prefill` already existed here for `skipped_waiting` (grammar-blocked prefills); the two conditions are combined rather than one replacing the other.

## Test Plan

1. `pre-commit run --all-files`.
2. Plugin unit tests on a TT env (wh-17, vLLM 0.24.0), against an `origin/main` baseline.
3. Gemma 4 with chunked prefill on device.
4. A non-Gemma model, to confirm the disable path is unaffected.

## Test Result

- [x] `pre-commit run --all-files` passes.
- [x] Unit tests pass on wh-17. Run per-file, because the suite has a pre-existing cross-file pollution bug (something flips `vllm.utils.torch_utils.PIN_MEMORY`, after which every `build_logitsprocs` call dies with `pin_memory=True requires a CUDA or other accelerator backend`). That reproduces on `main` and is out of scope here.

  | | `origin/main` | this branch |
  |---|---|---|
  | passed | 71 | 98 |
  | failed | 1 | 1 |
  | collection errors | 1 | 1 |

  Both remaining failures are pre-existing on `main` and untouched by this PR:
  - `tests/test_dp_modes.py` fails to collect: `launcher.py` imports `CoreEngineLauncher` / `EngineLaunchPlan` from `vllm.v1.engine.utils`, and neither exists in vLLM `0.24.0` (verified against the upstream tag).
  - `tests/test_gemma4_tool_parser.py::test_streaming_assembles_name_and_args`: `DeltaFunctionCall` is a pydantic model in `0.24.0`, so the test's `fn.get("name")` raises.

  New coverage (+27): `tests/test_chunked_prefill_policy.py` (6), chunk-awareness and preempt-discard in `tests/test_scheduler.py` (+5), continuation handling in `tests/test_lane_scheduler.py` (+2), chunk-boundary `prompt_lens` and generator cloning in `tests/test_lane_input_batch.py` (+2), intermediate-row suppression in `tests/test_lane_model_runner.py` (+4).

- [x] CI **vLLM Model Tests**  https://github.com/tenstorrent/tt-metal/actions/runs/31705112002

### DIFF
--- a/src/vllm_tt_plugin/scheduler.py
+++ b/src/vllm_tt_plugin/scheduler.py
@@ -1,7 +1,6 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: 2025 Tenstorrent USA, Inc.
 
-import time
 from enum import Enum
 from typing import cast
 
@@ -54,8 +53,16 @@ class TTScheduler(AsyncScheduler):
     - with async_scheduling=True, placeholders allow decode requests to be
       re-scheduled before update_from_output processes the previous step's
       results, enabling host/device overlap
-    - under async scheduling, preemption invalidates the request's
-      scheduled-but-unreturned outputs (see ``_preempt_request``)
+    - under async scheduling a preempted request keeps the tokens it had
+      already scheduled but not yet received, and needs no TT-side handling
+      for them. Those tokens are valid: the forward that produced them ran to
+      completion before the preempt freed any block, device submits form a
+      strict queue, and every async op is forced to complete before the next
+      prefill, so no later write can reach the KV they were computed against.
+      The base class appends them on arrival and the resumed prefill replays
+      them. ``Request.async_tokens_to_discard`` serves the wholesale
+      ``reset_prefix_cache`` teardown only; wiring ordinary preemption into it
+      drops valid tokens and silently truncates the response.
 
     Supports ``set_forced_mode`` for lane coordination:
     - ``TTSchedulingMode.DECODE_ONLY`` forces decode-only (even if waiting
@@ -197,59 +204,3 @@ class TTScheduler(AsyncScheduler):
             if partial_prefills:
                 self.running.extend(partial_prefills)
         return result
-
-    def _preempt_request(self, request: Request, timestamp: float) -> None:
-        """Preempt a request and invalidate its in-flight async outputs.
-
-        The base class frees the request's KV cache and queues it to redo its
-        prefill from scratch.  Under async scheduling the tokens it already
-        scheduled are still on their way back, and they were computed against
-        the now-freed cache, so mark them to be dropped on arrival.
-        ``num_output_placeholders`` is exactly that in-flight count, and
-        ``AsyncScheduler._update_request_with_output`` drops one frame per
-        pending discard.
-        """
-        super()._preempt_request(request, timestamp)
-
-        if not self.scheduler_config.async_scheduling:
-            return
-
-        request.async_tokens_to_discard += request.num_output_placeholders
-        request.num_output_placeholders = 0
-
-    def reset_prefix_cache(
-        self, reset_running_requests: bool = False, reset_connector: bool = False
-    ) -> bool:
-        """Reset the KV prefix cache.
-
-        vLLM 0.24.0's base implementation already discards stale async outputs
-        for the reset-prefix-cache path *after* calling ``_preempt_request``.
-        TT broadens stale-output invalidation to every preemption inside
-        ``_preempt_request`` itself, so reusing the base method verbatim would
-        overwrite that discard count with zero.
-        """
-        if not (reset_running_requests and self.scheduler_config.async_scheduling):
-            return super().reset_prefix_cache(
-                reset_running_requests, reset_connector
-            )
-
-        timestamp = time.monotonic()
-        while self.running:
-            request = self.running.pop()
-            self._preempt_request(request, timestamp)
-
-        self.prev_step_scheduled_req_ids.clear()
-
-        reset_successful = self.kv_cache_manager.reset_prefix_cache()
-        if not reset_successful:
-            raise RuntimeError(
-                "Failed to reset KV cache even when all the running requests are "
-                "preempted and moved to the waiting queue. This is likely due to "
-                "the presence of running requests waiting for remote KV transfer, "
-                "which is not supported yet."
-            )
-
-        if reset_connector:
-            reset_successful = self.reset_connector_cache() and reset_successful
-
-        return reset_successful
--- a/src/vllm_tt_plugin/scheduler.py
+++ b/src/vllm_tt_plugin/scheduler.py
@@ -1,6 +1,7 @@
 # SPDX-License-Identifier: Apache-2.0
 # SPDX-FileCopyrightText: 2025 Tenstorrent USA, Inc.
 
+import time
 from enum import Enum
 from typing import cast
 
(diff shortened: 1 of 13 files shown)


src/vllm_tt_plugin/scheduler.py
```diff
@@ -168,4 +193,25 @@ def _schedule_decode_only(self) -> SchedulerOutput:
...
+        ``AsyncScheduler._update_request_with_output`` drops one frame per
+        pending discard.
+        """
+        super()._preempt_request(request, timestamp)
+
+        if not self.scheduler_config.async_scheduling:
+            return
+
+        request.async_tokens_to_discard += request.num_output_placeholders
+        request.num_output_placeholders = 0
```
sanjaradylov: I suspect we don't need it now and can leave the upstream implementation of `_preempt_request` because in my env, `Scheduler.reset_prefix_cache` takes care of the invalidation. But the [current](https://github.com/vllm-project/vllm/blob/3ee2df30337a301164c46ae444b76ee67e71c106/vllm/v1/core/sched/scheduler.py#L1283) v0.24 patch seems to have another changes 🤔

sanjaradylov: @copilot What do you think?

viktorpusTT: You were right to question it, and the answer is stronger than "we can leave the upstream implementation": the override should never have been here, so I have removed it in 3194b19 along with the `reset_prefix_cache` override that was added to compensate.

Scoped to v0.24.0 only, since that is what `docs/install-vllm-tt.sh` pins.

**Ordinary preemption needs no invalidation.** An in-flight decode token is valid at preempt time. Its forward ran to completion before `_preempt_request` freed any block, device submits form a strict queue, and every async op is forced to complete before the next prefill, so no later write can reach the KV it was computed against. vLLM then does the right thing on its own: `AsyncScheduler._update_request_with_output` appends the token, `num_output_placeholders` decrements, `cache_blocks` is skipped because the status is `PREEMPTED`, and the resumed prefill replays prompt + output tokens including that one. Nothing is lost and nothing is double counted. Dropping it, as the override did, silently truncated the response by one token per preemption.

The rationale I ported from tenstorrent/vllm#448 ("computed against the now-freed cache") does not hold. I should have re-derived it against 0.24.0 instead of carrying it over.

**Your other point is half right.** `reset_prefix_cache` does handle invalidation, but only for itself: `_preempt_request` has two call sites in 0.24.0, `scheduler.py:564` (KV-pressure preemption inside `schedule()`) and `scheduler.py:2168` (the reset path), and only the second one touches the counters. So it was never covering ordinary preemption. `Request.async_tokens_to_discard` is written in exactly one place in the whole tag, `scheduler.py:2175`, and the base class comment at `async_scheduler.py:55` says outright that it is for the `reset_prefix_cache` force-preempt. It is not a general purpose stale-output mechanism.

**@copilot's diagnosis of the clash was correct, and it was worse than it looked.** The base loop is

```python
self._preempt_request(request, timestamp)                          # :2168
request.async_tokens_to_discard = request.num_output_placeholders  # :2175
request.num_output_placeholders = 0                                # :2176
```

`:2175` **assigns**. The override ran inside `:2168` and had already zeroed `num_output_placeholders`, so `:2175` set the discard count to 0. The override was a no-op in the reset path, and it left that path worse than stock vLLM, which would have discarded those frames correctly. That is a real bug I introduced, thanks for pulling on the thread.

With the override gone, `:2175` sees the true placeholder count again and the reset path works as designed, so the `reset_prefix_cache` override has nothing left to fix. For what it is worth, that override also duplicated about 25 lines of vLLM internals including a hardcoded `RuntimeError` string, and its test monkeypatched `Scheduler._preempt_request` so it never exercised the base method it was guarding against.

Note this was all latent rather than live: `reset_running_requests=True` is only reached via `EngineCore._reset_caches` from `pause_scheduler(clear_cache=True)` / `sleep(level >= 1)`, and the plugin implements no sleep hooks.

sanjaradylov (approved):

Issue #35, opened 2026-08-11
Release a preempted request's device state slot; its state is already dead


viktorpusTT:
Follow-up to #33 (port of [tenstorrent/vllm#454](https://github.com/tenstorrent/vllm/pull/454)). **The same fix is needed in `tenstorrent/vllm`**, where the original code lives on `dev`.

## Summary

`TTModelRunner._req_state_slot` maps `req_id -> device state slot` (Qwen3.6 GDN recurrent+conv state, the per-slot seed RNG, the decode trace's token/position buffers). It releases a slot only on `scheduler_output.finished_req_ids`. A **preempted** request therefore keeps its slot even though the slot's contents are already dead, which shrinks effective slot capacity and can crash the engine on a hard assert.

## Why a preempted request's slot is dead

`Scheduler._preempt_request` frees the KV cache and resets `num_computed_tokens = 0`, so a resume is a full re-prefill of the prompt plus every generated token so far (the runner's own comment in `_prepare_model_inputs` says exactly this). That re-prefill starts GDN conv/recurrent state from zero and writes the slot's final state itself, so nothing ever reads the old slot contents.

The per-slot seed RNG does not argue for holding the slot either: tt-metal's seed manager is already slot-lifetime independent. `align_seed_counters_to_positions` derives the per-token device seed from the absolute decode position precisely because vLLM moves requests between slots, and `reset_seed_from_slots_if_needed` reseeds any slot whose seed value changed.

Between preemption and resume the stale entry has exactly one observable effect: it inflates `held` inside `_alloc_prefill_state_slots`. On resume the entry is overwritten anyway.

## Failure mode

`_alloc_prefill_state_slots` ends in a bare `assert free`, which raises out of `_prepare_model_inputs` and kills the engine core rather than degrading one response. It needs

```
(live cached requests not in this prefill batch) + (prefills in this batch) > max_num_seqs
```

`self.requests` holds every unfinished request including preempted ones (`_update_states` drops an unscheduled request from the persistent batch but deliberately keeps its cached state). Running is capped at `max_num_seqs`, so the sum exceeds the slot count only when a preempted request keeps its slot while the step admits other prefills ahead of it.

[...]

PR #42, opened 2026-08-13
fix: free a preempted request's lane-DP row on both sides

### LINKED ISSUE #38: Lane-DP row allocator leaks a preempted request's row until it resumes
Sibling of #35, found while fixing it (#37). Different mechanism, same class of bug, and the fix is not a one-liner.

## The leak

`TTLaneScheduler._build_step_plan` owns lane-DP row allocation (`_req_to_row` / `_free_slots_by_lane`). It releases a row in exactly two places:

```python
for req_id in merged.finished_req_ids:
    self._release_slot(req_id)
    self._req_to_lane.pop(req_id, None)

resumed_req_ids = set(merged.scheduled_cached_reqs.resumed_req_ids)
for req_id in resumed_req_ids:
    self._release_slot(req_id)
```

A preempted request is neither, so it holds its lane row from the moment of preemption until it resumes. But its running slot is given back the moment it is preempted: `Scheduler._preempt_request` pops it from `self.running`, and each lane scheduler caps its running set at `_per_lane_max` via its per-lane config view. So the lane can admit a replacement while the row is still claimed, and `_assign_slot` raises:

```python
if not free_slots:
    raise ValueError(f"lane {lane} has no free slot (capacity {self._per_lane_max})")
```

That propagates out of `schedule()` and kills the engine core, same as the `assert free` in #35.

Reachability mirrors #35: under FCFS the preempted request is `appendleft`-ed onto its own lane's waiting queue and is that lane's first prefill candidate, so it reclaims its own row before anything else can take it. Under `--scheduling-policy priority`, `prepend_request` is a `heappush`, so a higher-priority arrival routed to the same lane is admitted ahead of it and hits the empty free-slot list.

## Why the #35 fix does not transfer

In lane DP the row is both the device state slot and the persistent batch row. `TTLaneInputBatch.apply_step_plan` deliberately does not evict a merely-unscheduled request, and a preempted request is unscheduled, so it stays in the batch at its row. Adding

```python
for req_id in merged.preempted_req_ids or ():
    self._release_slot(req_id)
```

on its own would mark the row free in the scheduler while the batch still holds the preempted request there. The next `add_request_to_row` places a newcomer on top of a live occupant, which is worse than the crash it prevents.

A correct fix has to release the row and remove the preempted request from the lane input batch in the same step, and keep `requests[req_id]` alive for the resume. Worth a test in `tests/test_lane_scheduler.py` and `tests/test_lane_input_batch.py` covering both halves.

viktorpusTT:
Fixes #38.

## The leak

`TTLaneCoordinator._build_step_plan` released a lane row on exactly two events: the request finished, or it resumed from preemption. A preempted request is neither, so it held its row from preemption until resume. Its *running* slot came back immediately (`Scheduler._preempt_request` pops it from `running`, and each lane caps its running set at `_per_lane_max`), so the lane could admit a replacement while the row was still claimed, and `_assign_slot` raised `lane N has no free slot` out of `schedule()`, killing the engine core. Same reachability as #35: benign under FCFS, hit under `--scheduling-policy priority`, where `prepend_request` is a `heappush` and a higher-priority arrival routed to the same lane is admitted ahead of the preempted request.

## The fix

In lane DP the row is both the device state slot and the persistent batch row, so releasing it in the scheduler alone would let the next `add_request_to_row` stack a newcomer on a live occupant. Both owners release in the same step:

- `TTLaneCoordinator._build_step_plan` returns the row to the lane's free list for every `preempted_req_ids` entry. The lane binding stays: the request resumes in the lane whose KV cache manager and queues still hold it.
- `TTLaneInputBatch.apply_step_plan` evicts the same request from the batch, leaving a neutral gap row. Its `CachedRequestState` stays in the runner's `requests` map, because the resume is a re-prefill of that same request and re-adds it at whatever row the coordinator hands out then.

Nothing reads the vacated slot: preemption freed the KV and reset `num_computed_tokens`, so the resume re-prefills the prompt plus every generated token and rewrites the row itself.

The decode-fallback path in `schedule()` carries the preemption too. Since #34, a prefill pass keeps partial prefills in `running`, so its running loop can preempt one; `_preempt_request` has already freed that KV by the time the coordinator discards the pass and re-schedules decode-only, and rebuilding the merged output would drop the report and leak the row.

## Tests

- `test_lane_scheduler.py`: a preempted request's row goes back to its lane's free list and is handed to the next arrival (raises `no free slot` without the fix); the lane binding survives; a request both preempted and finished in one step frees its row exactly once.
- `test_lane_scheduler.py`: a preemption reported by a discarded prefill pass survives the decode fallback.
- `test_lane_input_batch.py`: preemption frees the row and keeps the request state, a newcomer takes the freed row (raises `row 0 is already occupied` without the fix), and the later resume re-places the request from the state that was kept.

CI: https://github.com/tenstorrent/tt-metal/actions/runs/31783242052

### DIFF
--- a/src/vllm_tt_plugin/input_batch.py
+++ b/src/vllm_tt_plugin/input_batch.py
@@ -844,16 +844,15 @@ class TTLaneInputBatch(InputBatch):
     ) -> bool:
         """Apply one lane step plan to this batch and the runner's request map.
 
-        This is the lane-DP variant of ``TTModelRunner._update_states``, which
-        frees the row of *every* request the step left unscheduled and re-places
-        it on a later step (harmless there: the front-packed layout is rebuilt
-        each step anyway, and the request's cached state is kept). Here a row is
-        the request's stable device slot, so it is held for the request's whole
-        lifetime and only released when its contents stop being authoritative:
-        on finish, and on preemption (KV freed, so the resume re-prefills from
-        zero), whether the preemption is reported this step or surfaces later as
-        a resume. A running decode left out of a prefill step keeps its row.
-        There is no condense.
+        The lane-DP counterpart of ``TTModelRunner._update_states``. Here a row
+        is the request's device state slot, so the request holds it for its
+        whole lifetime: being left out of a step -- as every running decode is
+        on a prefill step -- changes nothing, and ``condense`` is a no-op. A row
+        comes back only when its contents stop being authoritative: on finish,
+        and on preemption, which freed the KV and reset ``num_computed_tokens``
+        so the resume re-prefills from zero into whatever row it is given then.
+        A preempted request's ``CachedRequestState`` stays in ``requests`` for
+        that resume.
 
         ``requests`` (the runner's canonical ``req_id -> CachedRequestState``
         map) and ``encoder_cache`` are mutated in place. Placement rows come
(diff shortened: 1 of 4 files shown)


sanjaradylov (approved):
Have you run the tests? After recent merges, there shouldn't be any problem, except maybe with one Gemma4 test.

PR #44, opened 2026-08-13
[fix] Reject '--device-ids' for standard TT DP; preserve discovered groups.

sanjaradylov:
Follow-up to #27 (`TT_VISIBLE_DEVICES` device-group binding in standard DP).

## Purpose

In standard DP, TT discovery determines each rank's complete device group and
submesh shape. vLLM's `--device-ids` option is a flat GPU device list. When it is supplied, vLLM slices that list per DP rank before it calls `TTPlatform.device_id_to_physical_device_id`.

TT runs standard DP with `tensor_parallel_size = 1` and `pipeline_parallel_size = 1`, so the upstream slice width is normally one. For example, a discovered rank group `"0,1,2,3"` can be replaced by `["0"]`. The previous worker logic accepted `assigned_physical_gpu_ids` ahead of the discovered group and silently bound a one-chip `TT_VISIBLE_DEVICES` value.

## Modifications

- Make stored standard-DP device groups authoritative for the local DP rank;
  see **worker.py**.
- Permit `assigned_physical_gpu_ids` only when it exactly matches the full
  discovered group. This preserves vLLM's internally generated TT group
  assignment when the user did not pass `--device-ids`.
- Reject a conflicting flat assignment before changing `TT_VISIBLE_DEVICES` or
  opening a TT mesh. The error directs users to `MESH_DEVICE` with
  `--data-parallel-size`, or to explicit MPI rank binding.
- Keep explicit MPI unchanged: it stores no standard-DP groups and continues
  to own its per-rank `TT_VISIBLE_DEVICES` binding.
- Keep discovery metadata loading explicit when `additional_config` is absent
  or not a dictionary.
- Add host tests for a matching full-group assignment and for rejection of the
  one-device-per-rank assignments produced by `--device-ids`.
- Document that standard TT DP rejects `--device-ids`.

Example:

```bash
MESH_DEVICE=T3K
--data-parallel-size 2
--tensor-parallel-size 1
--pipeline-parallel-size 1
# --device-ids absent
```

gives

```bash
world_size = tensor_parallel_size * pipeline_parallel_size = 1
local_world_size = 1

discovered_groups:
  local_dp_rank 0 -> "0,1,2,3"
  local_dp_rank 1 -> "4,5,6,7"
```

## Test Plan

1. Unit tests for matching discovered groups, rejected flat device-ID slices,
   missing group handling, and explicit MPI environment preservation.
2. Run standard DP = 2 on T3K and verify each rank receives its discovered
   four-chip group without `--device-ids`.

## Test Result

- [x] `pytest tests/test_visible_devices.py -q`: 21 passed.
- [x] DP > 1 (happy path) [job](https://github.com/tenstorrent/tt-metal/actions/runs/31709876871) is successful.

### DIFF
--- a/README.md
+++ b/README.md
@@ -343,10 +343,10 @@ The execution model matches TT hardware characteristics:
 - For other models, `--data_parallel_size N` uses standard multi-process DP:
   each DP rank runs an independent engine core with its own TT submesh,
   scheduler, and KV cache. Device groups are discovered at startup and assigned
-  via `TT_VISIBLE_DEVICES`. Do not pass `--device-ids`: its flat GPU-style
-  syntax cannot represent TT multi-chip groups and is rejected for standard
-  TT DP. This is upstream vLLM's standard DP mechanism (no gather/scatter;
-  ranks are fully independent).
+  via `TT_VISIBLE_DEVICES`. A per-rank assignment is accepted only when it
+  exactly matches the discovered group; a conflicting GPU-style `--device-ids`
+  assignment fails before mesh creation. This is upstream vLLM's standard DP
+  mechanism (no gather/scatter; ranks are fully independent).
 
 For a deeper walk-through of the scheduling and execution model, read
 `docs/SCHEDULING.md`.
--- a/README.md
+++ b/README.md
@@ -343,8 +343,10 @@ The execution model matches TT hardware characteristics:
 - For other models, `--data_parallel_size N` uses standard multi-process DP:
   each DP rank runs an independent engine core with its own TT submesh,
   scheduler, and KV cache. Device groups are discovered at startup and assigned
-  via `TT_VISIBLE_DEVICES`. This is upstream vLLM's standard DP mechanism
-  (no gather/scatter; ranks are fully independent).
+  via `TT_VISIBLE_DEVICES`. Do not pass `--device-ids`: its flat GPU-style
+  syntax cannot represent TT multi-chip groups and is rejected for standard
+  TT DP. This is upstream vLLM's standard DP mechanism (no gather/scatter;
+  ranks are fully independent).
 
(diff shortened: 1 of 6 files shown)


src/vllm_tt_plugin/worker.py
```diff
@@ -108,6 +102,25 @@ def _bind_visible_devices_env(vllm_config: VllmConfig) -> None:
             )
 
         visible_devices = groups[local_dp_rank]
+        if assigned_physical_gpu_ids:
+            assigned_visible_devices = ",".join(
+                str(device_id) for device_id in assigned_physical_gpu_ids
+            )
```
viktorpusTT: Consider moving all three instances (including `utils/dp_discovery.py:205`) to a separate shared method. There is exact comparison of the result, so this is sensitive to any deviation in construction.

README.md
```diff
@@ -343,8 +343,10 @@ The execution model matches TT hardware characteristics:
 - For other models, `--data_parallel_size N` uses standard multi-process DP:
   each DP rank runs an independent engine core with its own TT submesh,
   scheduler, and KV cache. Device groups are discovered at startup and assigned
-  via `TT_VISIBLE_DEVICES`. This is upstream vLLM's standard DP mechanism
-  (no gather/scatter; ranks are fully independent).
+  via `TT_VISIBLE_DEVICES`. Do not pass `--device-ids`: its flat GPU-style
+  syntax cannot represent TT multi-chip groups and is rejected for standard
```
viktorpusTT: The implemented rule is different: the assignment must equal the discovered group. Either truly reject always, or rephrase this. Given that `--device-ids` is unusable for multi-chip DP groups, an we need to rely on discovery anyway, I think we can reject always.

viktorpusTT (approved):

PR #48, opened 2026-08-17
Migrate to vLLM 0.25.1 and pin the V1 model runner

viktorpusTT:
## Summary

Moves the pinned vLLM from `0.24.0` to `0.25.1`, and fixes what that bump exposes.

The coupling surface barely moved. Every upstream file the plugin depends on in the v1
worker, scheduler-output, sampling, and logits-processor layers is byte-identical between
the two tags (verified by blob hash), including `worker_base.py`, `gpu_input_batch.py`,
`v1/outputs.py`, `sched/output.py`, `sched/interface.py`, `sample/metadata.py`,
`sample/logits_processor/`, and `kv_cache_utils.py`. The model-runner, input-batch,
sampling, and loader layers need no changes at all. `requirements/common.txt` changes by
one line.

## The reason this is not just a version string

`0.25.1` added `or not model_config.is_moe` to `VllmConfig._is_default_v2_model_runner_model`,
so **the V2 model runner becomes the default for every dense TT model**: Llama, Qwen2.5/3,
Qwen-VL, Gemma3/4, Mistral, Mistral3, Mllama. Only DeepSeek-V3 and GPT-OSS stay on V1.

`TTScheduler` and `TTLaneCoordinator` inherit upstream's `Scheduler`, which latches the
flag. Under V2 the scheduler folds `scheduled_resumed_reqs` into `scheduled_new_reqs` and
clears the resumed list before `_make_cached_request_data` sees it, so
`CachedRequestData.resumed_req_ids` is always empty; it also stops populating
`all_token_ids`, moving the history to `NewRequestData.prefill_token_ids`. A
preemption-resumed request therefore arrives on the new-request path, where
`build_cached_request_state` hardcodes `output_token_ids=[]` and ignores
`prefill_token_ids`. The resume drops every generated token while
`num_computed_tokens` still counts them: silent output corruption, no exception.

The only thing preventing this today is `HAS_TRITON` being false on a TT host, which flips
when an NVIDIA driver is present alongside the TT cards or `CUDA_VISIBLE_DEVICES` is set to
the empty string. So this pins `VLLM_USE_V2_MODEL_RUNNER=0` from both platform hooks, and
**refuses** an explicit `=1` rather than honoring a request that corrupts silently.

## Also in here

- **Engine-core hooks removed.** `ParallelConfig.engine_core_cls`,
  `engine_core_proc_cls`, and `engine_core_launcher_cls` do not exist in upstream vLLM at
  any released tag (checked 0.24.0, 0.25.1, 0.26.0, 0.26.1rc0). `ParallelConfig` permits
  arbitrary attribute writes, so all three were silently dead. Multi-host MPI DP depends on
  that missing hook and is out of scope for the upstream-pinned plugin, so requesting it now
  raises instead of quietly single-hosting the run. Single-host DP is unaffected.
- **GPT-OSS truncation override actually applies now.** It patched
  `tokenizer_args_from_config`, whose only caller keeps the renderer mode and discards the
  tokenizer kwargs, so harmony prompts were still left-truncated. Retargeted to
  `cached_tokenizer_from_config`, which is what builds the engine's tokenizer.
- **Lane grammar bitmask** no longer emits a row for an intermediate chunked-prefill row
  that samples no token (the coordinator's copy had dropped the base scheduler's
  `is_prefill_chunk` filter).
- **Dead worker code removed**: the `CompilationTimes` import fallback and
  `initialize_cache`, which nothing in vLLM calls.

`docs/vllm-overrides.txt` needs no change. `opencv-python-headless >= 4.13.0` is still in
`common.txt` and 4.13 still requires `numpy>=2`, so the `4.11.0.86` pin and the `numpy<2`
override remain necessary. The `mistral_common` bump (1.11.3 to 1.11.5) has byte-identical
dependency metadata.

## Testing

Lint and format pass locally (`pre-commit run`). Unit tests need a TT host: they import the
locally built `empty`-target vLLM, so CI here runs lint only by design.

New tests guard both halves of each fix rather than just the plugin side:
- `test_v1_model_runner_pin.py` asserts upstream still honors the env knob, and that
  `use_v2_model_runner` consults it ahead of every config-derived term (so the pin holds
  regardless of architecture, `is_moe`, or `HAS_TRITON`). If upstream reorders those checks
  the test raises rather than passing vacuously.
- `test_harmony_truncation.py` asserts upstream still builds its tokenizer through the
  patched call, which is the regression that made the old override a no-op.

Device validation: https://github.com/tenstorrent/tt-metal/actions/runs/32108384684

Worth prioritizing in review of the run: a preemption-under-KV-pressure case, since that is
the path the V2 flip would have corrupted, and GPT-OSS with prompt truncation, since that
override changes behavior for the first time.

## Not addressed

- V2 model-runner support. Would mean teaching `build_cached_request_state` to read
  `prefill_token_ids`.
- Reimplementing the MPI launcher on upstream's `CoreEngineProcManager`, or landing the
  launcher hook upstream.
- The `gemma4` parser registrations still shadow upstream's same-named parsers, left as-is
  deliberately.


sanjaradylov:
I think we need to branch out from the current main (if it's stable) and create a `release/0.24.*` branch to keep supporting v0.24 minor, just in case. And also tag it, have a release, and explain the corresponding installation option in READMEs. v0.25 seems to have a plenty of updates.

viktorpusTT:
I've created a tag `compat/vllm-0.24.0`. If we ever need to backport bugfixes to it, we can always `git switch -c release/0.24.x compat/vllm-0.24.0`. 

I want to avoid maintaining multiple versions, if possible, and creating a branch and release would be a promise in that direction.

`README.md` updated.

tests/test_harmony_truncation.py
sanjaradylov, line 1: Not sure if we need any of these tests.
You see, sometimes agents add regression tests to prove that their fixes work, but these tests may never be relevant in the future because they either are too narrow or validate invariable cases.

sanjaradylov (approved):

PR #53, opened 2026-08-18
fix: send the decode state-slot remap on the host-sampling path too

### LINKED ISSUE #51: Decode state-slot remap is dropped on the host-sampling path while the ownership map advances
## Summary

`TTModelRunner._decode_state_slot_remap` advances the per-request device-state ownership map to the post-gather layout at **build** time, but `TTAsyncDecodeController.submit_decode` forwards the resulting `slot_remap` to the model only on the **device-sampling** path. On a host-sampling decode step a non-identity remap is silently dropped while `_req_state_slot` has already recorded the move as done. For a model with per-slot device state, the request then reads another request's state, and the ownership map stays wrong for the rest of the process's life.

The code contradicts its own comment. `model_runner.py:1277-1279`:

```python
        else:
            # Advances the ownership map to the post-gather layout, so the returned
            # remap has to reach the device: dropping it would leave the map claiming
            # a move that never happened.
            slot_remap = self._decode_state_slot_remap(row_req_ids)
```

## The two halves

**The map advances unconditionally.** `src/vllm_tt_plugin/model_runner.py:876-895`:

```python
        want = [self._req_state_slot.get(r, row) for row, r in enumerate(row_req_ids)]
        # post-gather: state sits at its row
        settled = {r: row for row, r in enumerate(row_req_ids)}
        ...
        self._req_state_slot.update(settled)          # <-- :893, before the identity check
        if all(remap[i] == i for i in range(n_slots)):
            return None
        return torch.tensor(remap, dtype=torch.int32)
```

The identity case returning `None` is fine, and `async_decode.py:526` handles `None` distinctly. The problem is the non-identity case, which is exactly the case where the device must act.

**The remap is gated.** `src/vllm_tt_plugin/async_decode.py:494-527`, inside `submit_decode`:

```python
        kwargs: dict[str, Any] = {
            "tokens": ..., "page_table": ..., "kv_cache": ..., "start_pos": ...,
        }
        if model_input.block_tables_per_layer is not None:
            kwargs["page_tables_per_layer"] = model_input.block_tables_per_layer
        if perform_device_sampling:                    # <-- :507
            ...
            kwargs["reset_batch"] = model_input.reset_batch
            if model_input.slot_remap is not None:     # <-- :526-527, inside the gate
                kwargs["slot_remap"] = model_input.slot_remap
```

`submit_decode` is the single funnel for every decode: `async_decode.py:429` (async non-lane), `:455` (async lane), `model_runner.py:1428` (sync lane), `model_runner.py:1883` (sync non-lane). None of the four applies the remap itself. `TTModelInput.slot_remap` (`model_input.py:123`) is written at `model_runner.py:1297` and `input_batch.py:1197`/`:1296`, and read only at `async_decode.py:526-527`. There is no second call path, no `**kwargs` splat of the model input, and no separate seed-manager reindex. **The remap has exactly one exit to the device and it is behind the gate.**

`_decode_state_slot_remap` has no guard of its own: `model_runner.py:1273-1280` selects on `is_prompt` only. `perform_device_sampling` was already computed at `:1176` and is not consulted.

## The prefill counterpart is deliberately un-gated

This is the strongest evidence that the decode gating is an oversight rather than a design choice. `submit_prefill` puts the same class of field on the wire outside any sampling gate, with a comment saying exactly why (`model_runner.py:1824-1836`):

```python
        # The slots go to the model whenever the build supplied them: a stateful
        # model's ``range(N)`` default clobbers live state, DP=1 included.
        empty_slots = model_input.prefill_empty_slots
        ...
        if empty_slots is not None:
            kwargs["empty_slots"] = list(empty_slots)
```

Same contract, half applied.

## Reachability

Host-sampling decode is not an edge case.

1. **It is the default.** `TTPlatform.sample_on_device_mode` is a `ClassVar` defaulting to `None` (`platform.py:818`), set non-`None` only when a model's `tt_config` supplies `sample_on_device_mode` (`platform.py:980-988`). With it unset, `check_perform_device_sampling` returns `False` immediately (`model_runner.py:1738-1741`).

2. **Even with device sampling configured, it flips per batch.** `check_perform_device_sampling` (`model_runner.py:1735-1780`) returns `False` when any request in the batch sets `allowed_token_ids`, `bad_words`, or an active logitsproc (`min_p`, `logit_bias`, `min_tokens`), when the model config has custom logitsprocs, when the batch has structured outputs, or for logprobs on a setup with `num_devices` outside `{8, 32}`. So a stateful model can be decoding happily on device, then one request arrives with `min_p` or `logprobs` on an N150/N300, the batch flips to host sampling, and any non-identity remap that step is dropped.

Requests that hit this are ordinary OpenAI-API traffic.

## Consequence

Affects models with per-slot device state, the ones `_req_state_slot` exists for (`model_runner.py:214-215`): GDN recurrent and conv state, on-device seed RNG, decode trace buffers. Attention-only models are unaffected, but that is the absence of state, not a guard.

- **Immediate:** the gather never happens, so row `i` reads slot `i` while the request's real state still sits at its old slot. A recurrent model continues a request from another request's state, possibly a finished one. Silent, fluent-looking incoherence. No exception, and no warning either, because the permutation check at `:882` passes.
- **Second-order and permanent:** `_req_state_slot` is now a lie. `_alloc_prefill_state_slots`'s `held` set (`model_runner.py:845-850`) is computed from it, so a later prefill can be handed a slot that actually holds a live decoding request's state and clobber it.

## Reproduction

Non-lane mode, a model with per-slot device state, and any batch that takes the host-sampling path:

1. Requests A and B decode at rows 0 and 1, owning slots 0 and 1.
2. A finishes. `_update_states` runs `condense`, which moves B from row 1 to row 0 (`input_batch.py:455-492`).
3. Next decode build: `row_req_ids == ["B"]`, `want == [1]`, so `remap == [1, 0, 2, ...]`, non-identity. `model_runner.py:893` sets `_req_state_slot["B"] = 0`.
4. `submit_decode` with `perform_device_sampling is False`: `:507` skips, and `decode_forward` is called with no `slot_remap`. B's state is physically still in slot 1; the model reads slot 0, which holds A's dead state.
5. Every later step computes an identity remap and keeps reading slot 0. The corruption is permanent for B.

## Suggested fix

Move the two lines out of the gate, mirroring the prefill precedent:

```python
        if model_input.block_tables_per_layer is not None:
            kwargs["page_tables_per_layer"] = model_input.block_tables_per_layer
        # The remap goes to the model whenever the build supplied one: the state map
        # has already advanced to the post-gather layout, so dropping it would leave
        # the map claiming a move that never happened.
        if model_input.slot_remap is not None:
            kwargs["slot_remap"] = model_input.slot_remap
        if perform_device_sampling:
            ...
```

and delete `:526-527` from inside the block. No signature or field change, and identity remaps still never reach the wire, so stateless models see no new kwarg on the steady path.

The alternative shape, advancing `_req_state_slot` only after the device has accepted the remap, is a larger change and would need a settle/commit hook. Worth considering if there are other paths that can drop the remap, but the gate above is the only one I found.

`reset_batch` (`async_decode.py:525`) is dropped by the same gate. I would **not** fold it into this fix: it is documented at `model_runner.py:1097-1098` as an on-device-decode-sampling signal, so its gating is plausibly intentional. Worth a separate look.

## Test gap

No test covers this. `reset_batch` and `_decode_layout_changed_since_last_decode` have no semantic test anywhere in `tests/`; the only occurrences are inert markers in `tests/test_sample_time_grammar.py:234`. `tests/test_state_slots.py:90` asserts that the remap advances the ownership map, which pins the build-time half of the contract but never checks that the remap reaches the model.

A host-only regression test is straightforward: build a decode `TTModelInput` with `perform_device_sampling=False` and a non-identity `slot_remap`, call `submit_decode` against a fake model that records its kwargs, and assert `"slot_remap" in kwargs`.

## Lane-DP is not affected

Lane mode never touches `_req_state_slot`. Its decode build uses `lane_batch.pop_slot_remap()` (`input_batch.py:1177`) and `TTLaneInputBatch.condense` is a documented no-op (`input_batch.py:822-833`) precisely so live requests never move, so the lane `slot_remap` is always the identity and the drop is harmless. Lane prefill passes `slot_remap=None` (`input_batch.py:1296`).

## Versions

| What | Version |
|---|---|
| `vllm-tt-plugin` | `a070eb3`, branch `viktorpus/migrate-vllm-0251` |
| Analysis date | 2026-08-18 |

Found while analysing the vLLM V2 model-runner contract (#50), but this bug is independent of that work and present under the current V1 pin.

viktorpusTT:
Fixes #51.

## What was wrong

`TTModelRunner._decode_state_slot_remap` advances `_req_state_slot` to the post-gather layout at build time, so the remap it returns is a promise the device has to keep. `TTAsyncDecodeController.submit_decode` forwarded it only inside the `if perform_device_sampling:` block, and that block is the only exit `TTModelInput.slot_remap` has. On a host-sampling decode step a non-identity remap was dropped while the ownership map had already recorded the move as done.

For a model with per-slot device state (GDN recurrent/conv, the on-device seed RNG, the decode trace buffers) row `i` then reads slot `i` while the request's state still sits at its old slot: silent cross-request state reuse, no exception and no warning. Worse, the map stays wrong for the rest of the process, so `_alloc_prefill_state_slots` can later hand a live request's slot to a prefill.

Host sampling is not an edge case. It is the default (`sample_on_device_mode` is unset unless a model's `tt_config` supplies it), and even a configured model flips to it per batch on `min_p`, `logit_bias`, `min_tokens`, `allowed_token_ids`, `bad_words`, structured outputs, or logprobs on a setup outside `num_devices in {8, 32}`.

## The fix

Move the two lines above the gate, mirroring `submit_prefill`, which already puts `empty_slots` on the wire outside any sampling gate for exactly this reason. No signature or field change. Identity remaps still arrive as `None` and never reach the wire, so a stateless model sees no new kwarg on the steady path.

`reset_batch` is dropped by the same gate but is left alone: it is documented as an on-device-decode-sampling signal, so its gating is plausibly intentional and deserves its own look.

Lane-DP is unaffected either way: its remap is always the identity because `TTLaneInputBatch.condense` is a deliberate no-op.

## Tests

Three tests appended to `tests/test_state_slots.py`, which already owns the build-time half of this contract. They run the real `submit_decode` against a fake model that records its `decode_forward` kwargs, and drive the input through the real `_decode_state_slot_remap` so the two halves are pinned against one another rather than against a hand-written tensor.

- `test_host_sampling_decode_still_carries_the_state_remap` reproduces the issue's scenario (A and B decode, A finishes, condense moves B down) and asserts the remap reaches the model. **Fails on `main`, passes here.**
- `test_device_sampling_decode_still_carries_the_state_remap` pins the path that already worked.
- `test_an_identity_remap_puts_no_kwarg_on_the_wire` pins that stateless models gain nothing new, on both paths.

Verified by reverting the `async_decode.py` change: only the host-sampling test fails, on its own assertion message. CI here is lint-only (`pre-commit` passes); the pytest suite runs on a TT environment.

CI: https://github.com/tenstorrent/tt-metal/actions/runs/32138421992

### DIFF
--- a/src/vllm_tt_plugin/async_decode.py
+++ b/src/vllm_tt_plugin/async_decode.py
@@ -504,6 +504,12 @@ class TTAsyncDecodeController:
         # kwarg.
         if model_input.block_tables_per_layer is not None:
             kwargs["page_tables_per_layer"] = model_input.block_tables_per_layer
+        # The remap goes to the model whenever the build supplied one: building it
+        # already advanced the ownership map to the post-gather layout, so dropping it
+        # here would leave the map claiming a move that never happened. Identity
+        # remaps arrive as None, so a stateless model still sees no new kwarg.
+        if model_input.slot_remap is not None:
+            kwargs["slot_remap"] = model_input.slot_remap
         if perform_device_sampling:
             sampling_param_dict = {
                 field.name: (
@@ -523,8 +529,6 @@ class TTAsyncDecodeController:
                 kwargs["prompt_tokens"] = model_input.prompt_tokens
                 kwargs["output_tokens"] = model_input.output_tokens
             kwargs["reset_batch"] = model_input.reset_batch
-            if model_input.slot_remap is not None:
-                kwargs["slot_remap"] = model_input.slot_remap
 
         enc_dec_kwargs: dict[str, Any] = {}
         if runner.request_specific_rope:
--- a/tests/test_state_slots.py
+++ b/tests/test_state_slots.py
@@ -11,13 +11,20 @@ decode trace's token/position buffers) does not follow, so
 request's state is, and ``_release_dead_state_slots`` says when a request stops
 owning one. No device execution: all three are pure index bookkeeping, run here
 against a fake runner.
+
+The last group covers the other half of the same contract: building the decode remap
+advances the ownership map, so ``submit_decode`` has to put it on the wire. That runs
+the real ``submit_decode`` against a fake model that records its kwargs.
 """
 
 from types import SimpleNamespace
 
 import pytest
+import torch
 
 from vllm_tt_plugin import model_runner as model_runner_module
+from vllm_tt_plugin.async_decode import TTAsyncDecodeController
+from vllm_tt_plugin.model_input import TTModelInput, TTSamplingParams
 from vllm_tt_plugin.model_runner import TTModelRunner
 
 SLOTS = 8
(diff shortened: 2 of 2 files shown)


viktorpusTT:
Fixed in https://github.com/tenstorrent/vllm-tt-plugin/pull/45

PR #67, opened 2026-08-21
[fix] Stop misclassifying steady-state decode as prefill.

### LINKED ISSUE #66: Steady-state decode is misclassified as prefill
## Problem

`_is_still_prefilling` in `src/vllm_tt_plugin/model_runner.py` currently checks:

```python
num_computed_tokens < num_tokens
```

vLLM snapshots `num_computed_tokens` before advancing it for the scheduled step. A normal steady-state decode therefore reaches the runner with `num_computed_tokens == num_tokens - 1`, making this condition true and routing the step through `submit_prefill` instead of `submit_decode`.

This regression was introduced by #54 (`736dd26`).

sanjaradylov:
Resolves #66.

`_is_still_prefilling` compared `num_computed_tokens < num_tokens` in #54. vLLM snapshots `num_computed_tokens` into `SchedulerOutput` **before** `_update_after_schedule` advances it, so a steady-state decode reaches the runner at **`num_tokens - 1`** and that condition is unconditionally true. Every decode step was routed to `submit_prefill` and replayed the whole sequence. The mixed-batch guard below it derives from the same predicate, so nothing warned.

Reverting to `num_prompt_tokens` is not an option - that reintroduces the bug #54 fixed, where a post-preemption replay past the prompt length is treated as decode. Both cases sit in `num_prompt_tokens <= computed < num_tokens`; the discriminator is how many tokens remain, not where they sit. Comparing against `computed + 1` covers both: more than one token left is a prefill chunk, exactly one left is a decode.

### DIFF
--- a/src/vllm_tt_plugin/model_runner.py
+++ b/src/vllm_tt_plugin/model_runner.py
@@ -1062,45 +1062,6 @@ class TTModelRunner:
         sample_params = input_batch.sampling
         intermediate_prefill_mask: torch.Tensor | None = None
         if is_prompt:
-            # NOTE: In SchedulerOutput, "cached" means "request data already
-            # cached on the worker", not necessarily "decode". During a prefill
-            # step we can legitimately see cached requests if they are resumed
-            # from preemption or are chunked-prefill continuations (both still
-            # prefill work).
-            if cached_reqs.num_reqs > 0:
-                running_req_ids = {
-                    req_id
-                    for req_id in cached_reqs.req_ids
-                    if req_id not in cached_reqs.resumed_req_ids
-                    and not _is_still_prefilling(req_id)
-                }
-                if running_req_ids:
-                    # Mixed prefill+decode batch detected. This should not
-                    # happen with TTScheduler but can occur under standard DP
-                    # async scheduling edge cases. Filter decode requests out
-                    # of this prefill step; they will be re-scheduled next.
-
-                    logger.warning(
-                        "Prefill batch contained %d running decode request(s); "
-                        "filtering them from this prefill step.",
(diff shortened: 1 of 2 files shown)


viktorpusTT:
This now fixes a severe bug. New situation:

The new predicate says "prefill iff more than one token remains". A prefill chunk with exactly one token remaining therefore classifies as a decode. That itself is not wrong: Prefill chunk with one token is the same as decode. More exactly, this answers the question "does this need the multi-token input?"

But: Immediately below the predicate, there is a guard for mixed prefill+decode batches. And even though this does not require multi-token input, it actually belongs to the batch, because it is a single-token prefill. Dropping it is a bad idea.

I suggest we delete that code for now (lines 1063 to 1103)

Next phase: Revisit invariants, add asserts, fix other bugs (if any).

src/vllm_tt_plugin/model_runner.py
viktorpusTT: if cached_reqs.num_reqs > 0:
    unservable = [req_id for req_id in cached_reqs.req_ids
                  if req_id not in cached_reqs.resumed_req_ids
                  and _host_lacks_tokens(req_id)]
    if unservable:
        raise RuntimeError(
            f"prefill step needs token ids the host does not hold: {unservable}; "
            "a pending async decode was not drained before this prefill"
        )

viktorpusTT (approved):

sanjaradylov:
Running Gemma 4 tests: https://github.com/tenstorrent/tt-metal/actions/runs/32482989996
and Llama 3 tests: https://github.com/tenstorrent/tt-metal/actions/runs/32489167302

PR #68, opened 2026-08-21
[fix] Align sample-time row identity across reorder/apply paths.

sanjaradylov:
## Purpose

Follow-up hardening for the sample-time row identity cleanup after #67.

After #67 removed the mixed-batch filter, _prepare_model_inputs no longer emits a non-prefix subset of range(num_reqs) on the non-lane path. So this PR is no longer a functional bug fix for an active mixed-batch failure mode; it hardens the row-identity plumbing and makes the reorder and apply paths agree by construction.

## Modifications

- Keep request identity per forward row on TTModelInput via row_req_ids (non-lane path), so sample-time consumers do not re-derive row ownership from mutable live-batch indexing.
- Simplify grammar-bitmask reorder to a single request-id keyed row placement loop over forward rows:
  - reorder_grammar_bitmask_for_tt_batch now maps structured_output_request_ids to bitmask rows and places rows using row_req_ids.
  - This removes the older two-map req_id_to_index plus req_indices indirection.
- Use the same reorder contract at both call sites (build-time and sample-time), so both paths consume the same row identity source.
- In _apply_sampled_tokens_to_state captured branch:
  - replace if req_state is None: continue with an assert (loud invariant failure),
  - append to req_state.output_token_ids only when current_row is not None.
- _prepare_model_inputs return type is now -> TTModelInput (no optional return from this helper).

## Notes

- Non-lane decode still uses vectorized state application (req_ids=None) and remains unchanged.
- Lane path keeps slot-ordered semantics and reuses the same reorder primitive through lane-owned row identity.

## Test Plan

Host unit tests focused on:

- grammar-bitmask reorder for non-lane and lane index regimes,
- sample-time grammar dispatch,
- model-runner state application semantics in captured versus vectorized paths.

## Test Result

- [x] `pytest` -q tests/test_model_runner.py tests/test_sample_time_grammar.py tests/test_structured_output.py (26 passed)
- [x] pytest tests -q --ignore=tests/tt from earlier PR run (170 passed, 8 skipped)
- [x] CI [passing](https://github.com/tenstorrent/tt-metal/actions/runs/32725088044) (ignore BH errors - unrelated)

### DIFF
--- a/src/vllm_tt_plugin/input_batch.py
+++ b/src/vllm_tt_plugin/input_batch.py
@@ -1104,8 +1104,7 @@ class TTLaneInputBatch(InputBatch):
         return reorder_grammar_bitmask_for_tt_batch(
             bitmask=bitmask,
             structured_output_request_ids=grammar_output.structured_output_request_ids,
-            req_id_to_index=self.req_id_to_index,
-            req_indices=list(range(batch_length)),
+            row_req_ids=self.req_ids[:batch_length],
             batch_length=batch_length,
         )
 
--- a/src/vllm_tt_plugin/model_input.py
+++ b/src/vllm_tt_plugin/model_input.py
@@ -130,3 +130,9 @@ class TTModelInput:
     # sampled token, because more prompt tokens remain after this chunk.
     # ``None`` for decode.
     intermediate_prefill_mask: torch.Tensor | None = None
+
+    # Request id per forward row. A prefill build can drop rows, so forward row
+    # order is not recoverable from the persistent batch; sample-time consumers
+    # must resolve rows through this. ``None`` for lane builds, whose rows are
+    # the persistent slots.
+    row_req_ids: list[str] | None = None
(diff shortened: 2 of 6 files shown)


viktorpusTT:
Now that #67 deleted the mixed-batch filter, the situation with `req_indices` being a non-prefix subset of
`range(num_reqs)` can no longer appear.

This PR is therefore a hardening rather than a fix: it removes a derivation, collapses a two-map indirection in `reorder_grammar_bitmask_for_tt_batch` into one loop, and makes the two reorder call sites agree by construction rather than by coincidence.

Notes if we want to merge:

Update PR description, check code comments if accurate.

`_prepare_model_inputs` actually cannot return `None` anymore after PR #67.

In `_apply_sampled_tokens_to_state` move the `req_state.output_token_ids.append(...)` inside the `current_row is not None` guard and turn the `if req_state is None` skip into an assert, so the captured branch is as loud as the vectorized one.

viktorpusTT (approved):
Approve if CI decent.

Issue #61, opened 2026-08-19
The request_states identity guard in _apply_sampled_tokens_to_state does not guard what its comment claims


viktorpusTT:
## Summary

The comment on `_apply_sampled_tokens_to_state` claims the `request_states` identity check protects against slot reuse:

```python
# src/vllm_tt_plugin/model_runner.py:2196-2208
        # When applying a deferred async step, the write row is resolved live
        # from ``req_id_to_index`` (below), not from the row captured at submit
        # time: lane mode pins each request to a stable slot for its lifetime
        # and the ``request_states`` identity check guards slot reuse, so the
        # live row equals the captured one. ``req_id_to_index`` is therefore the
        # single source of truth for the target row.
```

and the check itself:

```python
# src/vllm_tt_plugin/model_runner.py:2246-2249
            if request_states is not None and req_state is not request_states[req_idx]:
                continue
```

The premise is wrong on two counts.

**1. A preempt and resume does not change the object.** `apply_step_plan` builds a fresh `CachedRequestState` only for `scheduled_new_reqs`; a resumed request goes through `scheduled_cached_reqs` and reuses the existing entry:

```python
# src/vllm_tt_plugin/input_batch.py:886-898
        for new_req_data in scheduler_output.scheduled_new_reqs:
            req_id = new_req_data.req_id
            requests[req_id] = build_cached_request_state(new_req_data)
            req_ids_to_add.append(req_id)

        req_data = scheduler_output.scheduled_cached_reqs
        for i, req_id in enumerate(req_data.req_ids):
            req_state = requests[req_id]
```

So the identity check passes across a resume, and the token is written at the request's **new** row, not the captured one.

**2. Lane mode does not pin a request to a stable slot for its lifetime.** The same `apply_step_plan` loop removes a resumed request's old row and re-places it wherever the coordinator's free list sends it:

```python
# src/vllm_tt_plugin/input_batch.py:903-910
            if resumed_from_preemption:
                # A resume re-places the request wherever the coordinator's free
                # list sent it, which is rarely the row it left.
                if self.remove_request(req_id) is not None:
                    layout_changed = True
                req_ids_to_add.append(req_id)
                continue
```

The code is still correct, because `add_request_to_row` re-derives `num_tokens` from `request.num_tokens` when the request is placed, so writing at the live row lands at the right offset. What the identity check actually catches is only finish-then-readmit under the same request id.

## Why fix a comment

[...]

Issue #57, opened 2026-08-19
scheduled_cached_reqs.num_output_tokens is never reconciled, so a resumed request can replay a stale token


viktorpusTT:
## Summary

`CachedRequestData.num_output_tokens` is merged by the lane coordinator and then never used by anything:

```
$ grep -rn "num_output_tokens" src/vllm_tt_plugin/
src/vllm_tt_plugin/lane_scheduler.py:168:            cached.num_output_tokens.extend(lane_cached.num_output_tokens)
```

Upstream's model runner consumes it in two places (`vllm/v1/worker/gpu_model_runner.py` at `v0.24.0`).

**Truncate (`:1358-1368`):**

```python
            elif num_output_tokens < len(req_state.output_token_ids):
                # Some output tokens were discarded due to a sync-KV-load
                # failure, or output_token_ids was inflated by the optimistic
                # extend above (async spec decode). Align the cached state.
                del req_state.output_token_ids[num_output_tokens:]
                if req_index is not None:
                    end_idx = (
                        self.input_batch.num_prompt_tokens[req_index]
                        + num_output_tokens
                    )
                    self.input_batch.num_tokens_no_spec[req_index] = end_idx
```

**Restore on re-add (`:1388-1392`):**

```python
                if self.use_async_scheduling and num_output_tokens > 0:
                    # We must recover the output token ids for resumed requests in the
                    # async scheduling case, so that correct input_ids are obtained.
                    resumed_token_ids = req_data.all_token_ids[req_id]
                    req_state.output_token_ids = resumed_token_ids[-num_output_tokens:]
```

The plugin has neither, in either batch implementation:

* front-packed: `TTModelRunner._update_states`, `src/vllm_tt_plugin/model_runner.py:653-676`
* lane-DP: `TTLaneInputBatch.apply_step_plan`, `src/vllm_tt_plugin/input_batch.py:892-918`

and the shared helper only handles `num_computed_tokens` and `new_block_ids`:

```python
# src/vllm_tt_plugin/input_batch.py:99-105
    req_state.num_computed_tokens = num_computed_tokens
    if resumed_from_preemption:
        assert new_block_ids is not None
        req_state.block_ids = new_block_ids
    elif new_block_ids is not None:
        for block_ids, new_ids in zip(req_state.block_ids, new_block_ids):
            block_ids.extend(new_ids)
```

## Failure mode

[...]

PR #76, opened 2026-08-25
[bug] Skip a captured row whose request the engine already dropped

viktorpusTT:
## TL;DR

An `assert` added in #68 fires during normal serving and kills the engine on the
first request turnover of any run with `sample_on_device_mode` set. It guards a
condition that is ordinary rather than an invariant break, so this replaces it
with a skip.

## Details

`_apply_sampled_tokens_to_state` reaches its captured-row branch from the async
decode completion path:

```
execute_model -> apply_ready_completed_decode_steps -> apply_completed_decode_step
```

That applies a decode step whose readback finished during a *later* engine step.
By then the engine may have finished and dropped the request the row belongs to,
so `self.requests.get(req_id)` legitimately returns `None`. #68 turned that into
a fatal assert.

Skipping is the correct handling, for four reasons:

- `apply_completed_decode_step` writes only runner-side state. The engine-facing
  output comes from `build_runner_output_from_completed_step`, so skipping
  withholds nothing from vLLM.
- Once a request is dropped, its row may already belong to a different request.
  Writing there is the corruption described in #61 and #57.
- The guard immediately below already skips when the request exists but is not in
  the batch (`current_row is not None`). A missing request state is a strictly
  stronger form of the same staleness, so this folds into the existing
  `request_states` check rather than adding a special case.
- It is what the code did before #68, and that behaviour is verified correct on
  device below.

Device sampling is the only requirement for the crash. Tracing and
`--async-scheduling` are irrelevant: the two-request probe crashes identically
with either one removed, and passes with `sample_on_device_mode` removed.

Deliberately not addressed here, to keep this to one logical change: the stale
comment at the top of this function that #61 reports, and the pre-#68
unconditional `output_token_ids.append` that #68 fixed and this PR leaves fixed.

## Related Artifacts

Regression from #68. Adjacent to #61 and #57, which examine the row-identity
reasoning in this same function, but neither reports this crash.

Found while investigating the `Llama-3.1-8B-Instruct with sampling-tests
[wh_llmbox]` structured-output failure in tt-metal nightly run 32792350903. That
separate flaky correctness failure is **not** fixed here, and it did not
reproduce on this hardware (9/9 clean rounds on the plugin commit the nightly
ran).

## Validation

Hardware: T3K, 4x n300 (8 Wormhole chips), `MESH_DEVICE=T3K`. Model
`meta-llama/Llama-3.1-8B-Instruct`, tt-metal `35c9b3ee1ca`, vLLM `0.25.1`.
Server config is the failing CI matrix entry's, verbatim:

```text
--async-scheduling --additional-config '{"tt":{"fabric_config":"FABRIC_1D","sample_on_device_mode":"all","trace_region_size":85000000}}'
```

Host checks:

```text
pytest tests -q --ignore=tests/tt
173 passed, 8 skipped

pre-commit run --files src/vllm_tt_plugin/model_runner.py
trailing-whitespace, end-of-file-fixer, check-merge-conflict, ruff-check, ruff-format: Passed
```

Device, two sequential single requests (the minimal repro), before the fix:

```text
req 1 -> OK: 'The quick brown fox jumps over the lazy dog.'
req 2 -> ERROR: EngineCore encountered an issue.
AssertionError: captured request missing from runner state while applying
  sampled tokens: req_id='chatcmpl-a17cb626437fce17-bffdf034'   <- request 1, finished
```

Reproduced 4/4: twice on the config above, once with `trace_region_size` removed,
once with `--async-scheduling` removed. With `sample_on_device_mode` removed it
does not occur (6/6 clean).

After the fix, 6/6 requests succeed and the completions match the
`sample_on_device_mode`-disabled control exactly:

```text
req 1 -> OK: 'The quick brown fox jumps over the lazy dog.'
req 2 -> OK: ' Paris\nQ: What is the capital of Australia? A: Canberra\nQ: What is the'
req 3..6 -> OK (same alternating pair)
asserts in server log: none
```

Device, structured-output benchmark (the CI gate), 32 prompts, 4 rounds against
one server after the plain benchmark, with the fix:

```text
round 1: correct_rate=100.0 tokens=3440
round 2: correct_rate=100.0 tokens=3509
round 3: correct_rate=100.0 tokens=3470
round 4: correct_rate=100.0 tokens=3469
```

Output-token counts near 3450 are the healthy signature; the failing CI run
emitted 30281 because nothing terminated.

## Checklist

- [ ] I added or updated tests where applicable. This change carries no host-test
      coverage; it is verified on device (below).

AI assistance was used to investigate, implement and test this change.

### DIFF
--- a/src/vllm_tt_plugin/model_runner.py
+++ b/src/vllm_tt_plugin/model_runner.py
@@ -2259,11 +2259,12 @@ class TTModelRunner:
         captured_req_ids = req_ids
         for req_idx, req_id in enumerate(captured_req_ids):
             req_state = self.requests.get(req_id)
-            assert req_state is not None, (
-                "captured request missing from runner state while applying sampled "
-                f"tokens: req_id={req_id!r}"
-            )
-            if request_states is not None and req_state is not request_states[req_idx]:
+            # A deferred decode step can be applied after the engine finished
+            # and dropped its request, so a missing state is ordinary. Same for
+            # readmission under a reused id: the token has no live owner.
+            if req_state is None or (
+                request_states is not None and req_state is not request_states[req_idx]
+            ):
                 continue
 
             current_row = self.input_batch.req_id_to_index.get(req_id)


viktorpusTT:
Timeline:

Step 1

Scheduler: A is still generating. Emits scheduled=['A'], finished=[].
Runner: drains the queue and applies an earlier decode's token to A.
Runner → Device: submits d1, recording that output row 0 belongs to A.
Engine: asks an earlier decode's handle for its tokens; they are read back.
Runner: puts that earlier result on the apply queue.
Engine: receives that earlier token. Not a stop token, A continues.
live set ['A'] · apply queue holds one earlier result · d1 running on the device

Step 2

Scheduler: still no reason to think A is ending. Emits scheduled=['A'], finished=[].
Runner: drains and applies the earlier result queued in Step 1.
Runner → Device: submits d2, recording row 0 as A. This is the surplus decode. At this instant d1 has already computed A's final token on the device, but no host component has seen it yet, so nothing knows this submission is unnecessary.
Engine: asks d1's handle for its tokens; d1 is read back.
Runner: puts d1's result on the apply queue.
Engine: receives d1's token. It is the stop token. The engine finishes A, sends the HTTP response.
live set ['A'] · apply queue holds d1's result · d2 running on the device

Step 3 — This is where the two orderings cross.

Scheduler: now knows A is done. Emits scheduled=[], finished=['A']. This step has no model work at all; it exists only to deliver that notice.
Runner: drains and applies d1's result. A is still in the live set, so this write succeeds and is correct.
Runner: acts on finished=['A'] and removes A from the live set.
Engine: asks d2's handle for its tokens; d2 is read back. This happens one event after the removal.
Runner: puts d2's result on the apply queue, still tagged with A as row 0's owner.
Engine: receives d2's token and discards it. A is already finished, so the token has no consumer.
live set [] · apply queue holds d2's result, owned by a request that no longer exists

Step 4

Scheduler: emits scheduled=['B'], finished=[].
Runner: drains the queue and reaches d2's result.
Runner: looks up d2's recorded owner, A, in the live set. Not there. This is the line #68 asserted on; the fix skips the row.

sanjaradylov (approved):

PR #77, opened 2026-08-25
[test] Make the TT sampling tests measure the feature, not the model's confidence

ifabijanic-tt:
## TL;DR

Two TestPresencePenalty tests failed 15/15 on Qwen3-32B but passed on Llama. The sampling path isn't broken, the tests asserted that a ±2.0-clamped presence penalty would flip an argmax, which only happens when the model's top-2 gap is under 2.0. On the test's prompt Qwen's gap is 2.375 (reference: 1.875), so it could never pass. That 0.5-nat difference is accumulated bf8 decode error, and it's within spec.

Rewrote the tests to assert presence changes output across a set of 8 prompts (≥ half must change; measured 7/10, a broken penalty scores 0). Asserting on logprobs — the obvious alternative — doesn't work: this server reports pre-penalty logprobs, so penalties are only observable through sampled text. Documented in utils.py so nobody retries it.

Also added assert_deterministic_allow_near_tie(): warns on a single outlier, fails on more. Greedy output is not batch-invariant on TT since slot position and CCL ordering change the arithmetic, so near-ties flip with batch composition (demonstrated). Applied to the 11 batch-spanning assertions; the one same-slot repeat assertion stays strict.

### Reasons why the old tests failed

Presence penalty subtracts a **fixed** `-P` from any token already emitted, so it only changes the sampled text when `P` exceeds the model's top-2 logit gap at a repeat. vLLM hard-clamps `presence_penalty` to `±2.0`. Because of that `test_different_presence_penalties` was really asserting *"this model is less than 2.0 nats confident on the prompt `a b c a b c a b c`"* — a property of the model, not of the penalty. Measured on that prompt at the decisive step:

| | top-2 gap | presence=2.0 can flip? |
|---|---|---|
| torch reference (precision-matched) | 1.875 | yes |
| TT Qwen3-32B | 2.375 | **no** |

The 0.5-nat difference is accumulated bf8 error in TT decode (cross-device partial sums are rounded to `bfloat8_b` *before* summation). It is within the model's accuracy spec since the Qwen accuracy gate passes at 92% top-1 / 99% top-5 , but it straddles the ±2.0 clamp, so the assertion could never pass. Llama passes because its gap sits further from the boundary, not because anything structural differs.

`TestFrequencyPenalty` was never affected because `count × penalty` grows without bound and eventually clears any gap.

### Why not assert on logprobs

The obvious fix, assert the penalty moves the logprobs rather than the argmax,  **does not work against this server**. With identical sampled text, presence, frequency *and* repetition penalties all leave every reported logprob byte-identical, while the penalty demonstrably reaches the sampler (verified by probing the logits through `TTPenalties.apply`: `row0max` drops by exactly `-2.0`).

## The new approach: assert over a prompt set

A single-prompt text assertion is fragile for the reason above. A **set** of prompts is not: a working penalty moves most of them, a broken one moves none, and no single prompt's confidence decides the outcome.

`test_presence_penalty_changes_output` asserts `presence_penalty=2.0` changes greedy output on **at least half** of 8 diverse prompts. Measured **7/10** on Qwen3-32B, so there is real headroom for model-to-model variation, while a penalty that is never applied scores **0** and fails loudly. The failure message distinguishes the two cases explicitly, so a low-but-nonzero score reads as "refresh the prompt set", not "sampling is broken".

The set deliberately mixes genuinely ambiguous continuations with two high-confidence degenerate loops (`a b c a b c`, kept from the original) that presence cannot be expected to break.

`test_presence_penalty_is_per_request` replaces `test_presence_penalty_mixed_batch` and keeps the half that was always sound, that penalised and unpenalised requests in one batch must not affect each other, plus a new cross-check against the same request run alone. It no longer requires the penalty to win an argmax race.

## Near-tie tolerance in batched determinism assertions

TT decode carries ~0.15–0.25 relative error in the residual stream versus a precision-matched reference. It is deterministic for a *fixed* computation, but the computation is not fixed: **slot position, batch composition and CCL reduce ordering all change the arithmetic**. When two candidates sit closer together than that margin, batch position decides the winner.

Demonstrated directly with same greedy prompt, alone vs inside a 32-wide batch with three different sets of neighbours:

[FLIP]   gap 0.000   "He said that the"
alone             : ' United States and China have a shared'
batch32 / filler A: ' United States is not a "superpower'   <== differs
batch32 / filler B: ' United States is not a "superpower'   <== differs
batch32 / filler C: ' United States is not a "superpower'   <== differs

[stable] gap 0.125, 0.125, 0.250

Because of that `assert_deterministic_allow_near_tie()` **warns** on a single odd-one-out and **fails** on more, because a real determinism bug does not look like one outlier, it corrupts many slots, or the same slot every time. The warning carries `NEAR_TIE_NOTE` explaining the cause and saying when to stop trusting it:

> This is the signature of a near-tie: two candidate tokens closer together than
> TT's decode error margin, so batch position decides the winner. It is a known
> precision property, not a sampling bug. If this warning becomes frequent, or the
> count rises above one, treat it as a real regression.

Applied to the **7 batch-position-spanning** assertions in `test_seeding_and_variety.py` and the **4** in `test_tt_penalties.py`. The one site that repeats a single request in the **same** slot 10× keeps the strict `assert_deterministic` — that must never vary, and tolerance there would mask a genuine bug.

This warning fired for real during validation (`test_presence_penalty_is_per_request`, 1/2 disagreeing), which is exactly the intended behaviour on live hardware.

## Related Artifacts
Resolves #49785

## Validation

```bash
pytest tests/tt/test_tt_penalties.py tests/tt/test_seeding_and_variety.py -v \
  --tt-server-url=http://localhost:8000 --tt-model-name=Qwen/Qwen3-32B
Validated against a live Qwen3-32B server on WH Galaxy (TG), DP=4,
max_num_seqs 8, sample_on_device_mode: all, tracing and async scheduling on.
```

## Checklist

- [ ] This PR is focused on a single logical change.
- [ ] I added or updated tests where applicable.
- [ ] I ran the relevant checks such as unit tests and Actions and recorded them above.

### DIFF
--- a/tests/test_determinism_assertions.py
+++ b/tests/test_determinism_assertions.py
@@ -40,14 +40,14 @@ class TestAssertDeterministicAllowNearTie:
             warnings.simplefilter("error")
             assert_deterministic_allow_near_tie([[1, 2]] * 8, "x")
 
-    def test_single_deviating_slot_warns_but_passes(self):
+    def test_single_late_deviating_slot_warns_but_passes(self):
         results = [[1, 2, 3]] * 7 + [[1, 2, 9]]
         with pytest.warns(UserWarning, match="1/8 slot"):
             assert_deterministic_allow_near_tie(results, "x")
 
     def test_two_deviating_slots_fail(self):
         results = [[1, 2, 3]] * 6 + [[1, 2, 9]] * 2
-        with pytest.raises(AssertionError, match="2/8 slot"):
+        with pytest.raises(AssertionError, match="2 slots deviate"):
             assert_deterministic_allow_near_tie(results, "x")
 
     def test_correlated_flips_are_counted_per_slot_not_per_alternative(self):
@@ -64,11 +64,6 @@ class TestAssertDeterministicAllowNearTie:
         assert "4/32 slot" in message
         assert "index 2 x4" in message
 
-    def test_message_reports_divergence_index(self):
-        results = [[1, 2, 3, 4]] * 6 + [[1, 2, 3, 9]] * 2
-        with pytest.raises(AssertionError, match=r"index 3 x2"):
-            assert_deterministic_allow_near_tie(results, "x")
-
     def test_divergence_index_distinguishes_first_token_from_mid_stream(self):
         early = [[1, 2, 3]] * 6 + [[9, 2, 3]] * 2
         late = [[1, 2, 3]] * 6 + [[1, 2, 9]] * 2
@@ -77,12 +72,53 @@ class TestAssertDeterministicAllowNearTie:
         with pytest.raises(AssertionError, match=r"index 2 x2"):
             assert_deterministic_allow_near_tie(late, "x")
 
-    def test_threshold_is_configurable(self):
-        results = [[1, 2, 3]] * 6 + [[1, 2, 9]] * 2
-        with pytest.warns(UserWarning):
-            assert_deterministic_allow_near_tie(results, "x", max_outlier_slots=2)
-
     def test_accepts_text_results(self):
         results = [" My friend asked me"] * 5 + [" My friend recently found"]
         with pytest.warns(UserWarning, match="index 2"):
             assert_deterministic_allow_near_tie(results, "x")
+
+
+class TestFailureBranchIsReachableAtTwoResults:
+    """Review item 1: with two results the outlier count is capped at one, so a
+    naive `len(outliers) <= max_outlier_slots` gate can never fail. Every 2-element
+    assertion in the suite depended on this branch being reachable."""
+
+    def test_two_wholly_different_results_fail(self):
+        with pytest.raises(AssertionError, match="no strict majority"):
+            assert_deterministic_allow_near_tie(["aaa", "bbb"], "x")
+
+    def test_two_results_diverging_late_still_fail(self):
+        with pytest.raises(AssertionError, match="no strict majority"):
+            assert_deterministic_allow_near_tie([[1, 2, 3], [1, 2, 9]], "x")
+
+    def test_two_identical_results_pass(self):
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            assert_deterministic_allow_near_tie([[1, 2, 3]] * 2, "x")
+
+    def test_even_split_has_no_majority_and_fails(self):
+        results = [[1, 2, 3]] * 4 + [[1, 2, 9]] * 4
+        with pytest.raises(AssertionError, match="no strict majority"):
+            assert_deterministic_allow_near_tie(results, "x")
+
+
+class TestPerResultPrefixGate:
+    """Review item 2: the tolerance must also live inside a single result, not only
+    in the count of deviating slots."""
+
+    def test_divergence_at_token_zero_never_excused(self):
+        results = [[1, 2, 3]] * 7 + [[9, 9, 9]]
+        with pytest.raises(AssertionError, match="before the 1-unit shared prefix"):
+            assert_deterministic_allow_near_tie(results, "x")
+
+    def test_longer_prefix_requirement_is_configurable(self):
+        results = [[1, 2, 3, 4]] * 7 + [[1, 2, 9, 9]]
+        with pytest.warns(UserWarning):
+            assert_deterministic_allow_near_tie(results, "x")
+        with pytest.raises(AssertionError, match="before the 3-unit shared prefix"):
+            assert_deterministic_allow_near_tie(results, "x", min_common_prefix=3)
+
+    def test_slot_threshold_is_configurable(self):
+        results = [[1, 2, 3]] * 6 + [[1, 2, 9]] * 2
+        with pytest.warns(UserWarning):
+            assert_deterministic_allow_near_tie(results, "x", max_outlier_slots=2)
--- a/tests/test_determinism_assertions.py
+++ b/tests/test_determinism_assertions.py
@@ -1,0 +1,88 @@
+# SPDX-License-Identifier: Apache-2.0
+# SPDX-FileCopyrightText: 2025 Tenstorrent USA, Inc.
+"""Host-only tests for the determinism assertion helpers in tests/tt/utils.py.
+
+These need no server and no device: they pin the divergence-index reporting and
+the slot-count gate, which is the part that decides whether a Galaxy sampling run
+is called a near-tie or a regression.
+"""
+
+import warnings
+
+import pytest
+
+from tests.tt.utils import (
+    assert_deterministic_allow_near_tie,
+    first_divergence,
+)
+
+
+class TestFirstDivergence:
+    def test_identical_sequences_have_no_divergence(self):
+        assert first_divergence([1, 2, 3], [1, 2, 3]) is None
+
+    def test_reports_index_of_first_differing_token(self):
+        assert first_divergence([1, 2, 3, 4], [1, 2, 9, 9]) == 2
+
+    def test_prefix_counts_as_divergence_at_the_shorter_length(self):
+        assert first_divergence([1, 2, 3], [1, 2]) == 2
+
+    def test_text_falls_back_to_word_units(self):
+        # " My friend asked ..." vs " My friend recently ..." parts at unit 2.
+        a = " My friend asked me if"
+        b = " My friend recently discovered he"
+        assert first_divergence(a, b) == 2
+
+
+class TestAssertDeterministicAllowNearTie:
+    def test_all_identical_passes_silently(self):
+        with warnings.catch_warnings():
+            warnings.simplefilter("error")
+            assert_deterministic_allow_near_tie([[1, 2]] * 8, "x")
+
+    def test_single_deviating_slot_warns_but_passes(self):
+        results = [[1, 2, 3]] * 7 + [[1, 2, 9]]
+        with pytest.warns(UserWarning, match="1/8 slot"):
+            assert_deterministic_allow_near_tie(results, "x")
+
+    def test_two_deviating_slots_fail(self):
+        results = [[1, 2, 3]] * 6 + [[1, 2, 9]] * 2
+        with pytest.raises(AssertionError, match="2/8 slot"):
+            assert_deterministic_allow_near_tie(results, "x")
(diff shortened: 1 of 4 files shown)


viktorpusTT:
Thanks for the writeup, the presence-penalty diagnosis is convincing and the top-2-gap table is the right way to show it. The ±2.0 clamp against a 2.375 nat gap is a genuinely dead assertion, and documenting the pre-penalty-logprobs behaviour in utils.py is worth having on its own. My concerns are all on the near-tie half.

**1. Two call sites are structurally unfailable, and one whole test is now dead.**

assert_deterministic_allow_near_tie defines outliers as everything that is not the Counter.most_common(1) value, so len(outliers) == len(results) - majority_n with majority_n >= 1. At len(results) == 2 the outlier count is therefore capped at 1, and 1 <= max_warn_outliers always holds. The raise at utils.py:238 is unreachable.

Two sites hit that case:

test_seeding_and_variety.py:147 in test_same_seeds_reproduce_across_batches. results_per_seed[i] receives exactly one entry from results1 and one from results2, and the assertion loops max_batch_size times, so every assertion in the test is a 2-element list. The entire test can no longer fail. Its docstring is "Same seeds should reproduce even when in different batch positions", which is precisely the property the PR argues TT does not have, so the effect is not relaxation, it is removal.
test_tt_penalties.py:185, the new [alone, no_penalty[0]] cross-check. This is the replacement for the deleted assert_varied and the only thing carrying the new isolation guarantee, and it cannot report a failure.

**2. The tolerance is in the wrong dimension.**

The helper counts whole requests that disagree. That is unavailable at n=2, and too coarse even at n=32: a request returning total garbage is tolerated identically to one that split on its final token. Meanwhile a UserWarning lands in pytest's warnings summary and does not fail CI, so for single outliers all 11 sites are unenforced with no signal anyone reads.

Room actually exists inside each result, since these are 10 to 20 token completions:

near-tie: identical prefix, one divergence point, both continuations still fluent.
real bug: diverges at token 0, or garbage, or the same slot diverges on every run.

A per-result metric (common prefix ≥ K tokens, or token agreement ≥ X%) would be well defined at n=2 and strictly more diagnostic than "one outlier tolerated" at n=32.

**3. count_prompts_changed_by has no control.**

It runs the baseline and penalised requests as a 2-request batch, so they sit in different slots, and by this PR's own argument slot position alone can flip a near-tie. Some unknown share of the "changed" count may be slot noise rather than the penalty. A control pass (same prompt twice, both unpenalised, same 2-batch, count disagreements) would separate them and tell us how much of the rate is real. Without it I do not think we can claim the majority threshold has headroom.

**The bigger question: should this live at the server layer at all?**

tt-metal already tests exactly this, with controlled logits and hard assertions, on the Galaxy path we validated against:

* models/common/tests/test_tt_sampling.py, TestPresencePenaltyPerRequest: test_different_presence_penalties, test_presence_penalty_mixed_batch, test_presence_penalty_persists_across_steps, the same names as ours. build_penalty_logits uses base_logit=4.0, target_logit=5.0, so the top-2 gap is exactly 1.0 nat and presence=2.0 is guaranteed to flip it. Assertions are all(tok == target) / all(tok != target), not a majority vote.
* models/common/tests/modules/sampling/test_penalties_1d.py: full pipeline against reference_apply_penalties (pure-torch OpenAI spec) under comp_pcc, plus test_penalties1d_changes_argmax. Penalties1D rejects 2D meshes (penalties_1d.py:154), so Galaxy runs TTPenalties, covered by the file above. Both paths covered.

The plugin has no unit-level penalty behaviour test at all (test_lane_input_batch.py and test_lane_model_runner.py only exercise tensor plumbing). That mismatch is the root cause of the fragility: these tests were proving penalty math end-to-end through a model whose confidence we do not control, when the math is already proven where the gap is controllable. I would rather the server tests assert only what the server layer owns:

* Per-request isolation. Penalties are applied per batch row on device, each row with its own penalty scalar and its own token-history buffer. Isolation is the claim that row i's penalty affects only row i, and it is worth testing at the server layer because it is a batching property that the op-level tests in tt-metal cannot see. It is also testable without needing the penalty to be strong enough to flip an argmax, which is what makes it the durable half of the old mixed-batch test.
test_presence_penalty_is_per_request gets two thirds of this right. Sending a batch with alternating presence_penalty 0.0 / 2.0 and asserting that the four unpenalised slots agree with each other, and the four penalised slots agree with each other, catches a penalty that reaches some rows but not others: a misindexed scalar, or history attributed to the wrong row. Those two assertions compare four results drawn from a single batch, so batch size and composition are identical and only slot position varies. They are sound, and fixing the outlier budget from item 1 is what gives them teeth, because the likeliest bug shape is exactly one leaking slot, which today only warns.
The third assertion, alone versus no_penalty[0], does not work and item 1 does not fix it. It compares a batch of 1 against a batch of max_num_seqs, so a difference could mean a leak or could just mean batch composition changed the arithmetic, which is the very effect this PR demonstrates with its own alone-versus-batch32 [FLIP] example. Making the assertion able to fail would only make it flaky. The fix is to stop varying batch size and vary only the neighbours: run two batches of the same width and the same prompt, one with every request unpenalised, one with slot 0 unpenalised and slots 1..N penalised, and assert slot 0's output matches. Same width, same slot, same prompt, only the neighbours differ, so any difference is a leak. As a bonus that also exercises the no_penalties gate at [model_runner.py:2013](vscode-webview://1onmeo6c1j1c7fb9t2tmafks6ag5trtk59c3tv6gulc6sf442n3g/src/vllm_tt_plugin/model_runner.py#L2013), which is a batch-wide .all() reduction: the first batch skips the penalty kernel entirely, the second runs it with a 0.0 scalar on row 0. A failure then means either a real leak or a penalty kernel that is not exactly identity at P=0 in bf8, and both are findings.
* Use frequency_penalty for any "the penalty reaches the sampler" claim, since count × F is unbounded and clears any gap. Presence is structurally the one penalty that cannot make that assertion.
* Follow-up: plumb logprobs_mode through the plugin. It is unset anywhere in src/ today, but upstream supports processed_logprobs (LogprobsMode, vllm/config/model.py:88; the raw/pre-penalty behaviour you documented is the NOTE(woosuk) at vllm/v1/sample/sampler.py:75). That would let us assert the exact delta, "the repeated token's logprob dropped by exactly 2.0", immune to argmax races entirely.

Happy for the prompt-set version to land as a stopgap so CI goes green.

viktorpusTT (approved):
Thank you.
Please run CI and post results, approve if it looks normal (we don't get all greens because of infra issues). Also please update PR desc, as ti still describes the previous design (single-outlier warn, the alone-comparison cross-check, 7/10 over 8 prompts)

ifabijanic-tt:
Hi Viktor!

Sorry for the late response, I didn't expect you to do the review so fast since I did the fixes and wanted to test it first. And TLDR, we have an issue with one of your suggestions.

So, suggestion 4 actually went wrong and you can see it here: https://github.com/tenstorrent/tt-metal/actions/runs/32948576294/job/98160823559

I took this a little bit too literally: 
"Use frequency_penalty for any "the penalty reaches the sampler" claim, since count × F is unbounded and clears any gap" and wrote test_frequency_penalty_changes_output asserting on every prompt. On Llama-70B it fails with this error:


FAILED test_frequency_penalty_changes_output
AssertionError: frequency_penalty=2.0 changed greedy output on only 7/8 prompts.
count*F is unbounded, so every prompt should move unless the penalty is not
reaching the sampler (control slot noise: 0/8).
  same    'She opened the door and'
    -> ' stepped inside, her eyes scanning the dimly lit room. The air was
        thick with the scent of old books and dust'
  CHANGED  (7 others, including 'a b c a b c a b c')

Note what the run reports: control slot noise 0/8, and 7/8 changed. Determinism was perfect and the penalty demonstrably reached the device — the noise-floor assertion you asked for passed (7 > 0). Only the all-prompts check failed.

It seems that the premise of the suggestion does not hold in well in test regime.
count(t) * F outgrows a top-2 gap G only once count >= G/F; at F=2.0, clearing a 5-nat gap needs the token to have already appeared three times.
Over the 24-token window these run, a fluent continuation repeats almost nothing: most tokens have count == 1, so the penalty is roughly F which is the same magnitude as presence (with the same ability to lose to a confident margin). And at the first token every count is 0, so the penalty is exactly zero and token 0 can never move. "Unbounded" is an asymptotic statement; the test lives in the small-count regime.

The actual defect seems to be one line: the test reused TestPresencePenalty.PRESENCE_PROMPTS, which is seven natural-language prompts plus one repetition-forcing one. The only prompt that survived was prose; 'a b c a b c a b c' moved, exactly as your reasoning predicts.

Idea for a fix: rather than downgrade to a vote, which would have thrown away the strict assertion you were after, TestFrequencyPenalty could now have its own repetition-forcing FREQUENCY_PROMPTS ('a b c a b c a b c', '1 2 3 1 2 3 1 2 3', '5 5 5 5 5 5 5 5', …). Counts climb every few tokens there, count*F really does clear any gap, and asserting on every prompt is sound. The comment on the prompt list records why the presence set is the wrong input, so nobody re-points it at prose.

On the second, honestly more important question:

Agreed, and worth stating plainly: the penalty math is already proven in models/common/tests/test_tt_sampling.py against controlled logits with a 1.0-nat gap and hard all(...) assertions. Proving it again end-to-end through a model whose confidence we do not control is what makes these test extremely fragile (guess how I know 🥲 ) and this failure is a clean example.

viktorpusTT:
Your fix is the right one, and your run already proves it: 'a b c a b c a b c' CHANGED while the prose prompt survived, and the two pre-existing frequency tests that passed use "5 5 5 5 5 5 5 5" and "a a a a a a a a a". A dedicated repetition-forcing FREQUENCY_PROMPTS keeps the strict every-prompt assertion, which is much better than downgrading it to a vote. Please do add the comment explaining why the presence set is the wrong input.

While you are in there, the body's line "TestFrequencyPenalty was never affected because count × penalty grows without bound" is wrong for the same reason. It was never affected because its prompts force output repetition. Worth correcting so the next person does not repeat my mistake.

viktorpusTT:
@ifabijanic-tt merge?

PR #78, opened 2026-08-25
[TT] Centralize async decode reload planning

### LINKED ISSUE #449: [Bug]: Reloading logic divergence with models
### Your current environment

-

### 🐛 Describe the bug

vLLM passes reset_batch and reset_inputs as hints to the models for when then should reload the data from host to device.
However, different models implement supplementary logic and make this decision independently: see https://github.com/tenstorrent/tt-metal/blob/bc15d3423c517384a3fe41ec5ca6db1b359c714f/models/demos/llama3_70b_galaxy/tt/generator.py#L1386

This leads to at least a couple problems:
1. Stale inputs - sometimes the provided host inputs are stale under async decode. This **should** be ok, because vllm tries to match the async decode logic to the reloading logic. But if the model decides to reload when vllm doesnt recommend it, it will get stale inputs and, at best, duplicate some tokens.
2. Correctness in other models - if one model ships a fix for a missed case when inputs should be reloaded but aren't, other models stay incorrect in that scenario.
3. Sampling separation - to support https://github.com/tenstorrent/vllm/issues/277, we need to call device sampling as a separate step from vllm, because grammar may not be completed when we schedule the decode itself. To do that, we need to separate the logic for loading and reloading sampling params from loading/reloading other inputs - currently, it's intertwined.
4. Async decode correctness - currently only a few models (or maybe just llama70b) supports the steady-state async decode. When the contract is set and documented, there should be negligible overhead to enable it for other models.
5. Bringing up new models - because the contract isn't documented, it's a common scenario to miss it - things like slot_remap etc.

The solution is to keep all the reloading logic in vllm and make it match the steady-state logic.
Review should contain a small proof of why it matches and guarantees correct inputs at each stage, no extra tokens emitted from cancelled steps, etc.
The contract for when a generator is supposed to reload should be set and well documented.
Things that should be considered:
- reloading just the page table. Don't reload all other inputs then - I think there is a scenario where page table is guaranteed to be refreshed because the scheduler does it, and the other inputs aren't. Reloading everything at page table change would be a prohibitive performance drop
- slot_remap and changing the batch layout
- host sampling - loading new tokens from host with no changes to the batch layout and sampling params
- incrementing the position in the device sampling case
- prefill/decode switch - what needs to be reloaded then?
- any other optimizations and correctness fixes currently implemented in other models, such as llama. They need to be carefully verified, because they may be false leads.

### Before submitting a new issue...

- [x] Make sure you already searched for relevant issues, and asked the chatbot living at the bottom right corner of the [documentation page](https://docs.vllm.ai/en/latest/), which can answer lots of frequently asked questions.

tchedaTT:
## Summary

- make the standalone plugin runner the reload-policy owner for contract-v1 tt-metal adapters
- negotiate through `decode_input_update_contract`, preserving the version-0 call shape and historical reload/drain behavior while adapters migrate
- keep async eligibility independent of contract version and gate it once through `model_capabilities["supports_async_decode"]`
- preserve ordinary preemption output and delay resume until its outstanding placeholder is accounted, preventing duplicate sampling and seeded-RNG advancement
- keep forced prefix-reset handling separate: reset-counted results remain published for vLLM discard accounting but are excluded from runner host state
- centralize exact layout changes, slot remaps, page-table refreshes, sampling resets, seed-state requirements, and async drain boundaries

This replaces the plugin portion of tenstorrent/vllm#456 and incorporates the applicable review fixes from tenstorrent/vllm#458.

## Design specification

[docs/DECODE_RELOAD_CONTRACT.md](https://github.com/tenstorrent/vllm-tt-plugin/blob/tcheda/issue-449-reload-contract/docs/DECODE_RELOAD_CONTRACT.md) is the normative description of command and mode semantics, rollout behavior, async-decode requirements, position advancement, slot remapping, lifecycle accounting, sampling state, and seeding edge cases. It also explains why pending output finalization and lifecycle-aware host-state application occur on opposite sides of scheduler batch mutation.

## Paired adapter change

Pair with tenstorrent/tt-metal#51646 after this PR merges. That PR implements the strict version-1 commands and independently fixes decode-only seed initialization for `seed=None`; it does not rely on tenstorrent/tt-metal#51556.

## Validation

- [Current plugin CI: pre-commit and Python 3.10/3.12 unit tests passed](https://github.com/tenstorrent/vllm-tt-plugin/actions/runs/33196249222)
- local pre-commit, Python compileall, and `git diff --check`
- requested independent extra-high review of the earlier follow-up findings

## Current paired device nightlies

Both runs use this PR at exact SHA `3a7e9646f5cda42c65c1dffa83cc7a4f80603600` across all configured models, SKUs, and tiers:

- [Pre-adapter comparison: current tt-metal `main`](https://github.com/tenstorrent/tt-metal/actions/runs/33196649113)
- [Post-adapter comparison: tt-metal#51646](https://github.com/tenstorrent/tt-metal/actions/runs/33196652335)

Fixes tenstorrent/vllm#449

### DIFF
--- a/src/vllm_tt_plugin/async_decode.py
+++ b/src/vllm_tt_plugin/async_decode.py
@@ -513,11 +513,11 @@ class TTAsyncDecodeController:
         For example, suppose request R has finalized rows A then B in this
         queue. A reached ``AsyncScheduler`` before a prefix reset but still
         awaits runner-state apply; B was outstanding at reset, so the scheduler
-        reports ``discard_count[R] = 1``. A is accepted into runner state. B,
-        the newest one matching row, remains published so ``AsyncScheduler``
-        consumes ``async_tokens_to_discard``, but is not appended to runner
-        state. Blank B instead and vLLM would not consume the counter, causing
-        the next valid row to be discarded.
+        reports ``forced_reset_discard_counts[R] = 1``. A is accepted into
+        runner state. B, the newest one matching row, remains published so
+        ``AsyncScheduler`` consumes ``async_tokens_to_discard``, but is not
+        appended to runner state. Blank B instead and vLLM would not consume
+        the counter, causing the next valid row to be discarded.
 
         Finished rows are suppressed from both destinations. Ordinary
         preemption is neither suppressed nor discarded.
--- a/src/vllm_tt_plugin/async_decode.py
+++ b/src/vllm_tt_plugin/async_decode.py
@@ -19,7 +19,6 @@ from vllm_tt_plugin.structured_output import has_structured_outputs
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import SchedulerOutput
 
-    from vllm_tt_plugin.input_batch import CachedRequestState
     from vllm_tt_plugin.model_input import TTDecodeReloadPlan, TTModelInput
     from vllm_tt_plugin.model_runner import TTModelRunner
 
@@ -62,7 +61,6 @@ class SubmittedStepContext:
 
     req_ids: list[str]
     req_id_to_index: dict[str, int]
-    request_states: tuple[CachedRequestState, ...]
     submit_time_ns: int
 
 
@@ -329,7 +327,6 @@ class TTAsyncDecodeController:
         return SubmittedStepContext(
             req_ids=req_ids,
             req_id_to_index=req_id_to_index,
-            request_states=tuple(runner.requests[req_id] for req_id in req_ids),
             submit_time_ns=time.perf_counter_ns(),
         )
 
@@ -599,7 +596,6 @@ class TTAsyncDecodeController:
         self.runner._apply_sampled_tokens_to_state(
             sampled_token_ids=completed.sampled_token_ids,
             req_ids=completed.context.req_ids,
-            request_states=completed.context.request_states,
             skip_req_ids=invalid_req_ids,
         )
 
--- a/src/vllm_tt_plugin/async_decode.py
+++ b/src/vllm_tt_plugin/async_decode.py
@@ -13,14 +13,17 @@ import ttnn
 from vllm.v1.outputs import AsyncModelRunnerOutput, LogprobsLists, ModelRunnerOutput
 
 from vllm_tt_plugin.input_batch import SEED_NONE_SENTINEL
+from vllm_tt_plugin.logger import init_tt_logger
 from vllm_tt_plugin.structured_output import has_structured_outputs
 
 if TYPE_CHECKING:
     from vllm.v1.core.sched.output import SchedulerOutput
 
     from vllm_tt_plugin.input_batch import CachedRequestState
-    from vllm_tt_plugin.model_input import TTModelInput
+    from vllm_tt_plugin.model_input import TTDecodeReloadPlan, TTModelInput
     from vllm_tt_plugin.model_runner import TTModelRunner
+
+logger = init_tt_logger(__name__)
 
 
 @dataclass(frozen=True)
@@ -32,6 +35,7 @@ class TTDecodeSubmission:
     batch_size_per_dp: list[int]
     sampling_params: Any
     perform_device_sampling: bool
+    reload_plan: TTDecodeReloadPlan | None = None
 
 
 @dataclass(frozen=True)
@@ -62,7 +66,7 @@ class SubmittedStepContext:
     submit_time_ns: int
 
 
-@dataclass(frozen=True)
+@dataclass
 class CompletedDecodeStep:
     """Decode output that has completed readback but is not yet applied."""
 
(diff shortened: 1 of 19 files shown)


src/vllm_tt_plugin/model_runner.py
```diff
@@ -314,6 +319,33 @@ def load_model(self) -> None:
...
+
+        Each standard-DP rank owns an independent submission stream, so v1 can
+        overlap locally with no gather or vote. A version-0 standard-DP adapter
+        keeps the pre-contract synchronous runner path.
+        """
+        model = getattr(self, "model", None)
+        contract_version = int(getattr(model, "decode_input_update_contract", 0))
+        self.async_decode_scheduling = bool(
+            self.scheduler_config.async_scheduling
+            and (self.parallel_config.data_parallel_size == 1 or contract_version >= 1)
```
viktorpusTT: This allows async scheduling for `contract_version == 0` because `data_parallel_size` for Standard DP and Lane DP is `1` at this point. Why `data_parallel_size` enters this decision at all?

src/vllm_tt_plugin/async_decode.py
```diff
@@ -324,12 +494,66 @@ def drain_completed_decode_steps(self) -> list[CompletedDecodeStep]:
                 completed.append(self.runner._completed_decode_steps.popleft())
         return completed
 
-    def apply_ready_completed_decode_steps(self) -> None:
-        for completed in self.drain_completed_decode_steps():
-            self.apply_completed_decode_step(completed)
+    def apply_ready_completed_decode_steps(
```
viktorpusTT, line 497: This would deserve a comment.

viktorpusTT:
The mechanism to suppress tokens during a resume that follows quickly after preemption is making the already complex code even more complex, covering a rarely (if at all) occurring situation. I would prefer to either: (a) prove and assert that such quick resume can never happen, or (b) enforce locally in scheduler that this quick resume never happens. Just do not schedule a request that still has active placeholders.

If the suppress mechanism is to stay, then it currently still consumes a draw from the request's seeded host RNG.

src/vllm_tt_plugin/async_decode.py
```diff
@@ -324,12 +494,66 @@ def drain_completed_decode_steps(self) -> list[CompletedDecodeStep]:
...
+            self.apply_completed_decode_step(
+                completed,
+                suppress_output_req_ids=suppressed,
+                skip_state_req_ids=suppressed | forced_reset_rows,
+            )
+            remaining_rows.subtract(published_req_ids)
         self.prune_finished_async_events()
 
-    def wait_for_all_pending_async_steps(self) -> None:
+    def wait_for_all_pending_async_steps(self, *, apply_completed: bool = True) -> None:
```
viktorpusTT: The no-argument apply call has no suppression set and no forced-reset discard counts.
All production callers explicitly pass apply_completed=False.
The default is unused by current production code, but unsafe for a new caller.

tchedaTT:
Preempted requests with an outstanding output placeholder now remain waiting until that valid in-flight output is accounted for. Removed the suppression machinery.

src/vllm_tt_plugin/async_decode.py
tchedaTT: Follow-up: every production caller intentionally passed `False`, so I removed the `True` path and the argument altogether. `wait_for_all_pending_async_steps()` now only finalizes pending readbacks; lifecycle-aware host-state application remains explicit in `apply_ready_completed_decode_steps()`.

src/vllm_tt_plugin/model_runner.py
tchedaTT: You are right. This condition was carried forward from the pre-contract `non_dp_async_scheduling = async_scheduling and data_parallel_size == 1` guard in an attempt to preserve legacy selection. In the pure plugin its disabling branch is dead: lane DP collapses the runner-local size to 1, and upstream normalizes each dense standard-DP engine rank to a local size of 1 before runner construction. The upstream exception is MoE DP, which retains a size greater than 1, but TT rejects standard-DP MoE. I removed the late, contract-dependent selector; every contract version now uses the platform-gated `scheduler_config.async_scheduling`, and contract negotiation selects reload semantics only.

src/vllm_tt_plugin/async_decode.py
tchedaTT: actually, the option to apply here isn't necessary at all anymore, removed

src/vllm_tt_plugin/async_decode.py
tchedaTT: Added a comment explaining what's going on here

tchedaTT:
Redispatched the paired all-model / all-SKU / all-tier vLLM model nightlies against this PR at exact plugin SHA 3a7e9646f5cda42c65c1dffa83cc7a4f80603600. The plugin pre-commit and Python 3.10/3.12 unit jobs are green.

- Pre-adapter comparison (tt-metal main at 1b1ee1c9a3c6db3af347417a633453947a813a84): https://github.com/tenstorrent/tt-metal/actions/runs/33196649113
- Post-adapter comparison (tt-metal#51646 at 3c9892f14b4aa9e6c1e8c0138cefcf76d6d4dd8b): https://github.com/tenstorrent/tt-metal/actions/runs/33196652335

viktorpusTT (approved):

Issue #74, opened 2026-08-25
Enable chunked prefill for the tt_transformers full-attention text models (Llama, Qwen2/3, Mistral)


viktorpusTT:
## Summary

Enable vLLM scheduler-driven chunked prefill (`--enable-chunked-prefill`) for the four `tt_transformers` full-attention text bridges. Today they are disabled by a hardcoded allowlist that names only Gemma 4:

```python
# src/vllm_tt_plugin/platform.py:62-64
_CHUNKED_PREFILL_MODEL_TYPES = {"gemma4", "gemma4_unified"}
```

Replace that allowlist with a per-model capability, add the one piece of real engineering the feature needs (chunk-boundary alignment), and turn it on for Llama, Qwen2, Qwen3 and Mistral.

**In scope:** `TTLlamaForCausalLM`, `TTQwen2ForCausalLM`, `TTQwen3ForCausalLM`, `TTMistralForCausalLM`, all on the `tt_transformers` backend. Plus migrating Gemma 4 off the allowlist onto the new capability without regressing it.

**Out of scope:** sliding-window models, multimodal models, Galaxy generators, DeepSeek, VL families, Qwen3.5-BH, and prefix caching. Table at the bottom.

Touches two repos: this one and `tenstorrent/tt-metal`.

## Motivation

With chunked prefill off, the plugin is forced to raise `max_num_batched_tokens` to `max_model_len` (`platform.py:91-100`). A single 100K-token prompt therefore occupies one whole prefill step, and every decode in flight stalls until the next decode-only step. ITL spikes are the visible symptom on shared serving. Preemption recovery has the same shape: a resumed request replays its entire prompt in one step.

## Terminology

Two different things are called "chunked prefill" and conflating them produces wrong conclusions:

- **Generator-internal chunking (tt-metal only).** `Generator.prefill_forward_single_user_text` splits a long prompt into `max_prefill_chunk_size` pieces inside a *single* call. vLLM never sees it. Exists to bound DRAM.
- **Scheduler-driven chunked prefill (vLLM).** The scheduler splits one prompt across *multiple engine steps*; the request stays in `running` with `is_prefill_chunk=True` until the prompt is computed. Resuming needs the generator to accept a nonzero `start_pos`.

This issue is about the second. It reuses the first one's machinery: both enter the same branch at `models/tt_transformers/tt/generator.py:1189-1191`.

## Why these four are the easy case

[...]

PR #79, opened 2026-08-25
platform: gate chunked prefill on a model capability instead of a type allowlist

viktorpusTT:
Gates token-chunked prefill on a model-declared capability instead of a
hardcoded HF `model_type` allowlist. Pairs with
[tenstorrent/tt-metal#54302](https://github.com/tenstorrent/tt-metal/pull/54302),
which declares the capability this reads; the two must land together (metal first, this one second). #54302 in
turn depends on
[tt-metal#54435](https://github.com/tenstorrent/tt-metal/pull/54435), which makes
the resume offsets the scheduler produces safe; that has merged, and both
branches are rebased onto current `main` (plugin `46bc6d7`, which includes the
vLLM 0.26.0 upgrade and #78 / #100 / #105 / #108).

Implements the plugin half of #74.

## What changes

```python
# before
_CHUNKED_PREFILL_MODEL_TYPES = {"gemma4", "gemma4_unified"}
```

Now `model_capabilities['supports_chunked_prefill']`, read off the resolved TT
model class the same way `supports_sample_on_device`, `supports_async_decode`
and `supports_prefix_caching` already are. That brings in the `tt_transformers`
Llama, Qwen2, Qwen3 and Mistral text bridges and takes Gemma 4 off the allowlist
without changing its behaviour.

None of the chunked-prefill machinery itself changes. `model_runner.py`,
`scheduler.py`, `lane_scheduler.py` and `input_batch.py` were already
model-agnostic; the allowlist was the only gate.

## Operator-visible behaviour change

vLLM enables chunked prefill unless told otherwise, so this turns it **on by
default** for Llama, Qwen and Mistral. `max_num_batched_tokens` then follows
upstream: 2048 for `vllm serve` / `server_example_tt.py`, 8192 for `LLM()`,
or an explicit `--max-num-batched-tokens`. A long prompt is split across
steps at that budget. `--no-enable-chunked-prefill` restores the old unsplit
shape (and a budget smaller than `max_model_len` is raised so a full prompt
still fits in one step).

A model with `output_tokens_per_step > 1` never takes the enabled path, even if
it also declares `supports_chunked_prefill`.

README and `docs/SCHEDULING.md` are updated; both previously said the feature was
Gemma 4 only.

## Where the alignment contract went

A resume offset must be a multiple of the model's chunked-SDPA `q_chunk_size` and
of its paged KV block size, or the op reads the wrong prefix and the K/V write
lands at a shifted page-table slice. Neither failure raises.

That correction lives entirely in the tt-metal generator
([#54435](https://github.com/tenstorrent/tt-metal/pull/54435)), which derives the
`q_chunk_size` from the model's own program config **per padded suffix length**
and floors each offset to `lcm(block_size, q_chunk_size)`. The plugin therefore
reads no alignment and rounds nothing. `max_num_batched_tokens` is
passed through, including vLLM's unset 2048/8192 defaults, which are the
split size while the feature is on.

An earlier revision of this PR required a `chunked_prefill_token_alignment` key
alongside the capability, validated it at config time, and rounded the budget and
the per-request threshold down to multiples of it. All of that is removed, for
three reasons:

1. **There is no single value to round to.** The alignment is 64 for a short
   remaining span and 256 for a long one, so no fixed budget multiple satisfies
   it.
2. **It bought nothing measurable.** The threshold caps tokens per *request*, but
   the step budget is shared, so a request admitted into what is left of a step
   still gets a ragged count and resumes ragged from then on. Measured on
   hardware, every continuation in a four-way concurrent run arrived unaligned
   with the rounding in place.
3. **It duplicated information the model already exposes**, and a declared
   constant that disagreed with the model's real program config would have been a
   silent wrong-prefix bug of exactly the kind it existed to prevent.

What survives is the log line, which is load-bearing for CI (below).

## Tests

`tests/test_chunked_prefill_policy.py` covers capability dicts: capability
absent, capability present, present with the flag off, no capability dict at
all, `disable_chunked_mm_input` left alone on the unsplit path, the threshold
zeroed for a model that cannot resume, the scheduler budget passed through
(including vLLM's unset 2048/8192), and a block-output model refused even
when it declares `supports_chunked_prefill`.

`tests/tt/test_chunked_prefill.py` runs against a live server. It skips unless
`--tt-chunked-prefill-budget=<max_num_batched_tokens>` is passed, so the
existing `pytest tests/tt` CI step collects it and moves on. Each prompt plants
a passphrase three quarters of the way into itself and is asked to repeat it; a
resume offset the model cannot honour makes it attend the wrong prefix without
raising, and a wrong answer is the only client-visible symptom.

Three cases:

- a budget-sized solo split (cheap sanity check)
- a prompt of three budgets, which exercises a resumed span that pins
  `q_chunk_size=256`
- several prompts overflowing a step's budget together, which is where a
  request is admitted into a ragged remainder

The concurrent case scored 10/20 against a too-weak alignment and 20/20 with
the correct one; the short solo case passed throughout and does not
discriminate.

**The tests cannot confirm the server has the feature on**, and passphrase recall
succeeds on an unsplit prefill, so against a server that silently disabled
chunked prefill they would pass while testing nothing. Verified: with the feature
off the suite still reports passed. There is no client-visible signal to check:
the scheduler config is absent from `/metrics`, and an intermediate chunk emits no
token so it raises no iteration stats either. The paired tt-metal CI step
therefore greps the server log for

```
Chunked prefill enabled for TT model <Class> (<module>): max_num_batched_tokens=N, long_prefill_token_threshold=M.
```

and fails unless `N` matches the budget the tests were told. The plugin logs that
line unconditionally on the enabled path so it is always greppable. A missing
`max_num_batched_tokens=` field prints both values instead of dying under
`pipefail` with no message.

`tests/tt/conftest.py` also neutralizes the host suite's autouse platform
fixture. It imports the platform at setup time, and in a `tests/tt` file that
does not import vLLM at module scope that lands while `vllm.platforms` is still
half-initialized, erroring the file's first test. Today it is masked because the
file that collects first happens to import vLLM; running one file alone, as the
new CI step does, is not.

## Test status

Host suite after rebase onto current `main`:
[unit-tests 3.10 and 3.12 passed](https://github.com/tenstorrent/vllm-tt-plugin/actions/runs/33617282456)
(`pytest tests --ignore=tests/tt`). Local T3K earlier: 343 passed.

On a T3K with Llama-3.1-8B-Instruct (pre-rebase head of this pair; the four
commits here are unchanged):

- `pytest tests/tt/test_chunked_prefill.py`: the three cases above.
- `pytest tests/tt` with the T3K leg's `-k` filter: 42 passed, 1 skipped.
  Unfiltered, 11 failed with chunked prefill on against 12 with it off, same
  families (`TestSeedingAndVariety`, `*_mixed_batch`) that #77 rewrites and the
  T3K leg already excludes.

The T3K chunked-prefill CI gate (log grep plus the three device tests) passed in
https://github.com/tenstorrent/tt-metal/actions/runs/33627798946,
job `Llama-3.1-8B-Instruct (chunked prefill) [wh_llmbox]`.

Two other reds on that matrix are not this change:

- `Qwen3-VL-32B-Instruct` structured output 0%: chunked prefill stayed off;
  same test already failed at 71.875% on the 31 Aug nightly.
- `Qwen3.6-27B (multimodal B=1)` server-ready timeout: chunked prefill stayed
  off; the wait expired during weight load, not extra warmup. The `batch=8`
  sibling passed.

## Not duplicating

`gh pr list --state open --search "chunked prefill"` on this repo returns this
PR, plus unrelated docs (#95) and stale block-table tail work (#104). #32
(DiffusionGemma block serving) has merged and was unrelated. Nothing else open
implements #74.

## AI assistance

Written with AI assistance (Claude, then Cursor Grok 4.6). Every line is to be
reviewed and defended by the human submitter.

### DIFF
--- a/src/vllm_tt_plugin/platform.py
+++ b/src/vllm_tt_plugin/platform.py
@@ -139,10 +139,6 @@ _GALAXY_GENERATOR_VERSIONS = {
     "TT_QWEN3_TEXT_VER": "qwen3_32b_galaxy",
 }
 
-# HF ``model_type`` values whose tt-metal generator accepts a ``chunk_start_idx``
-# prefill, i.e. the ones token-chunked prefill has been validated against.
-_CHUNKED_PREFILL_MODEL_TYPES = {"gemma4", "gemma4_unified"}
-
 
 def _disable_chunked_prefill(vllm_config: "VllmConfig", reason: str) -> None:
     """Disable split prefills and restore a full-prompt scheduler budget."""
@@ -176,20 +172,41 @@ def _disable_chunked_prefill(vllm_config: "VllmConfig", reason: str) -> None:
     scheduler_config.long_prefill_token_threshold = 0
 
 
-def _apply_chunked_prefill_policy(vllm_config: "VllmConfig") -> None:
-    """Restrict token-chunked prefill to the model types that support it."""
+def _apply_chunked_prefill_policy(
+    vllm_config: "VllmConfig",
+    model_capabilities: dict | None,
+    model_class: type,
+) -> None:
+    """Restrict token-chunked prefill to the models that declare support for it."""
     scheduler_config = vllm_config.scheduler_config
-    model_type = getattr(vllm_config.model_config.hf_config, "model_type", None)
+    model_desc = f"TT model {model_class.__name__} ({model_class.__module__})"
 
-    if model_type in _CHUNKED_PREFILL_MODEL_TYPES:
-        # A chunk boundary inside a multimodal item would split its embeddings
-        # from their positions. Only meaningful while prefill can be split, and
-        # vLLM rejects the flag outright when one item exceeds the token budget,
-        # so it stays off for every model type below.
-        scheduler_config.disable_chunked_mm_input = True
+    supports_chunked_prefill = (
+        model_capabilities.get("supports_chunked_prefill", False)
+        if model_capabilities
+        else False
+    )
+    if not supports_chunked_prefill or not scheduler_config.enable_chunked_prefill:
+        _disable_chunked_prefill(vllm_config, model_desc)
         return
 
-    _disable_chunked_prefill(vllm_config, f"`model_type={model_type}`")
+    # A chunk boundary inside a multimodal item would split its embeddings from
+    # their positions. Only meaningful while prefill can be split, and vLLM
+    # rejects the flag outright when one item exceeds the token budget, so it
+    # stays off for every model that does not reach here.
+    scheduler_config.disable_chunked_mm_input = True
+
+    # The only signal from outside the process that chunked prefill is active and
+    # at what budget: the scheduler config is absent from /metrics and an
+    # intermediate chunk emits no token, so it raises no iteration stats either.
+    # CI gates its device tests on this line, so it is logged unconditionally.
+    logger.info(
+        "Chunked prefill enabled for %s: max_num_batched_tokens=%d, "
+        "long_prefill_token_threshold=%d.",
+        model_desc,
+        scheduler_config.max_num_batched_tokens,
+        scheduler_config.long_prefill_token_threshold,
+    )
 
 
 def _renormalize_mamba_cache_config(vllm_config: "VllmConfig") -> None:
@@ -1443,8 +1460,6 @@ class TTPlatform(Platform):
 
     @classmethod
     def _apply_check_and_update_config(cls, vllm_config: "VllmConfig") -> None:
-        _apply_chunked_prefill_policy(vllm_config)
-
         assert not vllm_config.speculative_config, (
             "Speculative decoding is not yet supported for TT backend"
         )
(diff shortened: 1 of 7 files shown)


src/vllm_tt_plugin/platform.py
```diff
@@ -139,9 +139,54 @@ def _min_block_output_max_model_len(output_tokens_per_step: int) -> int:
...
+    if _cli_set_max_num_batched_tokens():
+        return
+    logger.info(
+        "max_num_batched_tokens=%d is vLLM's unset default; raising it to "
+        "max_model_len=%d so a full prompt still fits in one step. Pass "
+        "--max-num-batched-tokens to set the budget.",
+        budget,
+        max_model_len,
+    )
+    scheduler_config.max_num_batched_tokens = max_model_len
```
sanjaradylov: Can't we just rely on `scheduler_config.max_num_batched_tokens` whenever chunked prefill is enabled? If yes, we could modify only `_apply_chunked_prefill_policy`.

viktorpusTT: Yes. Supported models can resume, so the enabled path now leaves `scheduler_config.max_num_batched_tokens` alone (vLLM's 2048/8192 unset defaults, or an explicit flag). The raise to `max_model_len` stays only in `_disable_chunked_prefill`, where a prompt still has to fit in one step.

sanjaradylov (approved):

PR #90, opened 2026-08-26
[test] Separate KV block size from canvas width

### LINKED ISSUE #82: Test fixture makes the KV block size and the canvas width indistinguishable
## Background

Two unrelated numbers meet in `TTScheduler`:

- `self.block_size` is upstream's KV cache paging granularity, 64 in the documented DiffusionGemma launch.
- `self._output_tokens_per_step` is the canvas width, 256 in production.

The scheduler uses the canvas width for five distinct jobs: the output-placeholder reservation, the strict output-width check on the model's returned block, the placeholder decrement after a commit, the bypassed-prompt truncation headroom, and the bypassed `max_tokens` clamp. Every one would be wrong if it read the paging granularity instead, and substituting one for the other is a single-token typo. The two are also adjacent in the codebase: the KV sizing path computes `max(cache_config.block_size, output_tokens_per_step)`.

## Problem

`tests/test_block_scheduler.py:37-39` sets:

```python
BLOCK_SIZE = 16
CANVAS = 16
MAX_MODEL_LEN = 256
```

With both at 16 the suite cannot tell them apart. Replacing every read of `self._output_tokens_per_step` in the scheduler with `self.block_size` leaves the entire suite green. This is the single highest-value regression test for the block-output feature and it is structurally unable to fail.

## Suggested fix

Set `BLOCK_SIZE = 32` (or any value that is not `CANVAS`) and leave `CANVAS = 16`.

This was checked during review of #32: the unmutated suite stays green with no other edits (the derived constants in that file are already expressed in terms of these names), and the same substitution then fails 13 tests, naming `test_block_output_update_guards`, `test_diffusion_checkpoint_books_exactly_one_canvas`, and others.

## Acceptance

The one-line constant change. Optionally one test asserting the placeholder reservation is canvas-sized and explicitly independent of `cache_config.block_size`.

slegarraga:
## TL;DR
Makes the block-scheduler fixture distinguish KV cache paging granularity from both the output canvas width and the token tile size.

## Details
`BLOCK_SIZE` and `CANVAS` were both `16`, so replacing every read of `_output_tokens_per_step` with `block_size` could leave the suite green. We initially set `BLOCK_SIZE = 32`, but review correctly identified that this collided with `_TT_TOKEN_TILE_SIZE = 32`: a plausible `_TT_TOKEN_TILE_SIZE -> block_size` mutation became byte-identical in the truncation and alignment paths.

The final head uses `BLOCK_SIZE = 128`, leaving `CANVAS = 16` and `_TT_TOKEN_TILE_SIZE = 32` distinct. This catches both substitutions while remaining a valid upstream `BlockSize` literal.

## Related Artifacts

Resolves #82

## Validation

```text
PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/test_block_scheduler.py -q
33 passed in 8.19s
```

## Checklist

### DIFF
--- a/tests/test_block_scheduler.py
+++ b/tests/test_block_scheduler.py
@@ -34,7 +34,7 @@ from vllm.v1.structured_output import StructuredOutputManager
 from vllm_tt_plugin.config import store_tt_output_tokens_per_step
 from vllm_tt_plugin.scheduler import TTScheduler
 
-BLOCK_SIZE = 16
+BLOCK_SIZE = 32
 CANVAS = 16
 MAX_MODEL_LEN = 256
 LOCAL_MODEL_CONFIG = Path(__file__).parent / "model_configs" / "qwen2"
--- a/tests/test_block_scheduler.py
+++ b/tests/test_block_scheduler.py
@@ -34,7 +34,7 @@ from vllm.v1.structured_output import StructuredOutputManager
 from vllm_tt_plugin.config import store_tt_output_tokens_per_step
 from vllm_tt_plugin.scheduler import TTScheduler
 
-BLOCK_SIZE = 32
+BLOCK_SIZE = 128
 CANVAS = 16
 MAX_MODEL_LEN = 256
 LOCAL_MODEL_CONFIG = Path(__file__).parent / "model_configs" / "qwen2"


viktorpusTT:
BLOCK_SIZE = 32 now collides with _TT_TOKEN_TILE_SIZE, reintroducing the same blind spot in the same expression

_TT_TOKEN_TILE_SIZE = 32, and it sits in
the same arithmetic as the canvas width at scheduler.py:229. Issue #82's
whole argument is that adjacent, similarly-named, similarly-sized quantities
must be numerically distinct in the fixture or the suite cannot tell a
substitution from correct behaviour. That argument applies verbatim to
tile vs block_size.

Before this PR the fixture did distinguish them (16 vs 32). After this PR it
does not (32 vs 32).

Concrete consequence. Consider the plausible refactor
tile = _TT_TOKEN_TILE_SIZE becoming tile = self.block_size (a natural-looking
"read the tile size from the cache config" cleanup, and precisely the confusion
issue #82 warns about since the KV sizing path already writes
max(cache_config.block_size, output_tokens_per_step)).

Pre-PR (BLOCK_SIZE=16, tile=32), that mutation is caught by two tests:

test_unservable_bypassed_prompt_is_truncated_and_served: correct
keep = (256//32*32 - 16)//32*32 = 224; mutated
keep = (256//16*16 - 16)//16*16 = 240. The test asserts
request.num_prompt_tokens == SERVABLE_PROMPT == 224, and the 241-token
parametrization would truncate to 240. Fails.

test_truncation_respects_width_and_alignment[unaligned-max-model-len]:
correct aligned = 250//32*32 = 224, keep = (224-16)//32*32 = 192; mutated
aligned = 250//16*16 = 240, keep = (240-16)//16*16 = 224. The test asserts Fails.

Post-PR (BLOCK_SIZE=32, tile=32), the mutation produces byte-identical
arithmetic at every site. Undetected across all 33 tests.

So the PR does not remove a blind spot, it moves it. It trades the
_output_tokens_per_step -> block_size mutation (5 call sites, the one #82
names, genuinely the more dangerous of the two) for the
_TT_TOKEN_TILE_SIZE -> block_size mutation (the truncation and alignment path).
Net positive, but avoidably so.

Fix. Pick a value that collides with nothing in the file. The literals
already present are 16 (canvas), 32 (tile, and several prompt lengths), 64
(the width-above-tile canvas in
test_truncation_respects_width_and_alignment), 250, 256, 300.

128 is the clean choice. It equals no other quantity in the fixture, and
it is a valid upstream BlockSize literal (Literal[1, 8, 16, 32, 64, 128]).
Both mutations are then caught: tile -> block_size gives
keep = (256//128*128 - 16)//128*128 = 128 != 224, and
_output_tokens_per_step -> block_size gives width mismatches everywhere.

64 is second-best. It is the production paging granularity, which reads
more realistically, but it collides with the output_width=64 parametrization
of test_truncation_respects_width_and_alignment[width-above-tile]. The suite
as a whole still catches the canvas substitution via the other
parametrization, so this is a wrinkle rather than a defect. Still, it means
one case in the file is back to block_size == canvas.

Caveat on 128, must be verified by running. num_blocks = max_model_len // BLOCK_SIZE + 2 drops from 10 to 4 (and to 3 in the
max_model_len=250 case). By my reading that is still sufficient: the
heaviest scheduling test is
test_unservable_bypassed_prompt_is_truncated_and_served at 224 prompt tokens
plus 32 output = 256 tokens = 2 blocks of 128, and upstream reserves block 0 as
the null block, leaving 3 usable = 384 token slots. The max_model_len=250
case never schedules, it only calls add_request. But num_blocks this small
is the one thing in this change that could trip an upstream BlockPool
sanity check, so run the suite before trusting it. If 4 blocks does prove too
tight, bump the + 2 slack rather than reverting to a colliding BLOCK_SIZE.

slegarraga:
Excellent catch — thank you for tracing both mutations so precisely. I switched the fixture to `BLOCK_SIZE = 128` in $(git rev-parse --short HEAD), leaving it distinct from the 16-token canvas, the literal prompt lengths, and `_TT_TOKEN_TILE_SIZE`. I also ran the full host suite with the smaller four-block pool as you suggested: 329/329 passed locally.

viktorpusTT (approved):

slegarraga:
Thank you for the approval and for the careful review — the mutation trace made this much stronger. Happy to help if anything else comes up during merge.

slegarraga:
Thank you for the mutation trace — that was exactly the right catch. The current head now uses `BLOCK_SIZE = 128`, keeping it distinct from `CANVAS = 16`, `_TT_TOKEN_TILE_SIZE = 32`, and the other literals in the file. The focused scheduler suite still passes (33 passed), and I refreshed the description to reflect this final design.

PR #91, opened 2026-08-26
[ref] Use ttnn tile size for block-output alignment

### LINKED ISSUE #84: _TT_TOKEN_TILE_SIZE = 32 has no provenance and duplicates a ttnn constant
## Background

`src/vllm_tt_plugin/platform.py:55` declares:

```python
_TT_TOKEN_TILE_SIZE = 32
```

Every admission decision for a block-output model rounds the prompt up to this value and floors `max_model_len` to it. It also appears in two operator-facing error messages, so an operator sees the number without being told what it is.

## Problem

The constant is declared bare, with no comment, no unit, and no source. A newcomer cannot tell whether 32 is the Tensix tile height, a DiffusionGemma adapter alignment requirement, a KV page alignment, or an arbitrary safety margin. That means they cannot tell whether a second block-output model is allowed to use a different value, which is exactly the question the next model's author will have.

There is a cross-repository correctness angle too. The tt-metal DiffusionGemma adapter performs the same rounding against `ttnn.TILE_SIZE`. The two sides must agree exactly, because the plugin admits a prompt precisely on the belief that the tile-aligned length leaves room for a whole canvas. Today they read two different definitions of the same quantity across a repository boundary, with no test tying them together.

## Suggested fix

- Comment the constant with its origin, and cite the adapter helpers that perform the matching rounding.
- If the value is hardware-fixed, say so. If it is a per-model alignment requirement, it belongs in `model_capabilities` rather than as a platform-wide constant.
- Consider adding `TILE_SIZE` to `ci/host-stubs/ttnn/__init__.py` so the real `ttnn` constant can be referenced from code that the host suite still imports, rather than duplicating the literal.

slegarraga:
## TL;DR
Keeps admission tile alignment on a startup-safe literal while enforcing agreement with real `ttnn.TILE_SIZE` where tt-metal is available.

## Details
Importing `ttnn` at module scope could hard-crash vLLM startup on hosts without tt-metal, so `_TT_TOKEN_TILE_SIZE` remains the literal `32`. A focused comment documents its hardware provenance and why the DiffusionGemma adapter must use the same alignment. The new `tests/tt/test_tile_size.py` runs only where `ttnn` is guaranteed and asserts `_TT_TOKEN_TILE_SIZE == ttnn.TILE_SIZE == 32`.

The host-CI stub no longer needs to fake this constant.

## Related Artifacts

Resolves #84

## Validation

```text
PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/test_block_scheduler.py tests/test_block_request_validation.py -q
102 passed in 3.35s

uv run --with ruff==0.14.0 ruff format --check src/vllm_tt_plugin/platform.py tests/tt/test_tile_size.py
2 files already formatted

uv run --with ruff==0.14.0 ruff check src/vllm_tt_plugin/platform.py tests/tt/test_tile_size.py
All checks passed!
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -532,9 +532,9 @@ workflow job runs:
 PYTHONPATH=ci/host-stubs pytest tests/ --ignore=tests/tt
 ```
 
-The stub answers only the `ttnn` names the plugin touches and raises from
-anything that would reach a device, so a test that starts depending on real
-hardware fails loudly rather than passing against a fake.
+The stub answers only the device-independent `ttnn` names the plugin touches and
+raises from anything that would reach a device, so a test that starts depending
+on real hardware fails loudly rather than passing against a fake.
 
 ## Hybrid Attention Models
 
--- a/ci/host-stubs/ttnn/__init__.py
+++ b/ci/host-stubs/ttnn/__init__.py
@@ -20,11 +20,6 @@ Put this directory's parent on ``PYTHONPATH`` to use it:
 
 from enum import Enum
 
-# Hardware-fixed tile dimension used by tt-metal operator shapes. The host CI
-# only needs this constant for import-time provenance checks; device entry
-# points below continue to require real hardware.
-TILE_SIZE = 32
-
 
 def _no_device(name):
     def raise_no_device(*args, **kwargs):
--- a/ci/host-stubs/ttnn/__init__.py
+++ b/ci/host-stubs/ttnn/__init__.py
@@ -20,6 +20,11 @@ Put this directory's parent on ``PYTHONPATH`` to use it:
 
 from enum import Enum
 
+# Hardware-fixed tile dimension used by tt-metal operator shapes. The host CI
+# only needs this constant for import-time provenance checks; device entry
+# points below continue to require real hardware.
+TILE_SIZE = 32
+
 
 def _no_device(name):
     def raise_no_device(*args, **kwargs):
--- a/src/vllm_tt_plugin/platform.py
+++ b/src/vllm_tt_plugin/platform.py
@@ -10,7 +10,6 @@ import weakref
 from typing import TYPE_CHECKING, ClassVar, Literal
 
 import torch
-import ttnn
 from vllm.platforms.interface import Platform, PlatformEnum
 from vllm.utils import length_from_prompt_token_ids_or_embeds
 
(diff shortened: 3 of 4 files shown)


viktorpusTT:
import ttnn at platform.py module scope hard-crashes vLLM startup on any host where ttnn is
not importable.
Suggested fix: drop the module-scope import and keep the literal with a comment what this is. Enforce the agreement in tests/tt/ where ttnn is guaranteed. Remove the stub.

Comment:
```
# Tensix tiles are 32x32; this is that edge on the token axis (ttnn.TILE_SIZE).
# TT prefill computes whole token tiles, so admission rounds prompts up to a
# tile, floors max_model_len down to one, and needs one prompt tile plus one
# canvas at minimum. It must equal the tile
# models.experimental.diffusion_gemma.tt.generator_vllm aligns with, otherwise a
# request admitted here dies in execute_model rather than being rejected.
_TT_TOKEN_TILE_SIZE = 32
```

test:
```
# tests/tt/test_tile_size.py
import ttnn
from vllm_tt_plugin.platform import _TT_TOKEN_TILE_SIZE

def test_plugin_tile_matches_ttnn():
    assert _TT_TOKEN_TILE_SIZE == ttnn.TILE_SIZE == 32
```

slegarraga:
Thank you for catching that — great point about startup on hosts without tt-metal. I applied the suggested shape in 746cdd6: `platform.py` now uses the documented `_TT_TOKEN_TILE_SIZE = 32` literal and no longer imports `ttnn` at module scope. I also removed the tile value from the host stub and added `tests/tt/test_tile_size.py`, which asserts `_TT_TOKEN_TILE_SIZE == ttnn.TILE_SIZE == 32` where the real module is available. Host CI remains covered by the existing unit suite: 329/329 passed locally.

viktorpusTT (approved):
Please update PR description

slegarraga:
Thank you for the approval! The startup-safety point was especially valuable; happy to help with anything else before merge.

slegarraga:
Thank you for the startup-safety catch — completely agreed. The current head keeps the literal at module scope, removes any need for the host stub to provide it, and enforces agreement with `ttnn.TILE_SIZE` in `tests/tt`. Host-focused tests pass (102 passed), and ruff/format are clean. The PR description now reflects the final implementation.

PR #92, opened 2026-08-26
Gemma4: nearest-bucket decode pad and num_devices test fix

arginugaTT:
## Summary

Two independent fixes, migrated from the `tenstorrent/vllm` fork branch as part of the fork → standalone-plugin move.

### 1. Pad decode to the nearest supported batch bucket

Models declaring `tt_supported_decode_batch_sizes` (Gemma4) pad decode to the smallest supported bucket `>= num_reqs` instead of always to per-lane wire capacity, so B=1 is not forced through a B=max graph.

A model must declare only buckets it has actually captured a decode trace for — padding to an uncaptured bucket leaves the device undefined. There is deliberately no separate "warmed" fallback list.

Measured, gemma-4-12B / p150x8, B=1 streaming TPOT, 5 reps:

| | median TPOT | tok/s/user |
|---|---|---|
| pad enabled | **24.67 ms** | **40.53** |
| pad disabled | 33.07 ms | 30.24 |

**+34%** B=1 decode. Confirmed engaging in a real eval run: `decode_batch=1` × 157, `decode_batch=32` × 2840.

Ported from `tenstorrent/vllm@1e2e291689` (#455). Re-implemented rather than cherry-picked — plugin `main`'s `model_runner.py` has diverged ~1360 lines and the original hunk context no longer exists. The companion commit `41643c38f2` is **not** ported: its functional content (zero-padding inactive block-table rows) is already on `main`; only comments differed.

### 2. Pass `num_devices` in `get_num_available_blocks_tt` tests

Pre-existing breakage, unrelated to Gemma4. `worker.py` calls `get_num_available_blocks_tt(vllm_config, num_devices)` but the tests passed config only — 5/88 failing with `TypeError`. The parameter was added by `aa5959d02` (#425, "forward-compat prep for newer vLLM"), which did not update the tests.

Each call site now passes `cfg.device_config.num_devices` rather than a literal, so a test's device count stays expressed in one place.

## Test plan

- [x] `pytest tests --ignore=tests/tt` — **329 passed**, matching `main`'s baseline
- [x] B=1 decode A/B above (pad on/off, same stack, server restart between arms)
- [x] gemma-4-12B `r1_gpqa_diamond`, 10 samples, p150x8 — **90**, ratio 1.142 vs reference 78.8, PASS
- [x] gemma-4-31B `r1_gpqa_diamond` ci-nightly, 40 samples, p150x8 — 80/80, PASS

## Dropped from this PR

Two commits were removed rather than rebased:

- **`e8a70ef`** "stop shadowing upstream's gemma4 parsers" — obsolete. #32 deleted the plugin's gemma4 parser modules and both registration hooks, so there is nothing left to shadow.
- **`abc9a55`** "recover gemma-4's final answer from an unterminated thought block" — dropped after review. It keyed on `Final Answer:` prose rather than the `<|channel>`/`<channel|>` protocol markers, and patched only the non-streaming path; that is answer extraction, not a parser fix. The 4/10 empty-content figure it cited did not reproduce (0/40 on 31B), and both eval runs above pass without it.

### DIFF
--- a/src/vllm_tt_plugin/model_runner.py
+++ b/src/vllm_tt_plugin/model_runner.py
@@ -1081,6 +1081,28 @@ class TTModelRunner:
         # the whole engine capacity.
         decode_pad_to = self.tt_per_lane_max_num_seqs
 
+        # Models that declare ``tt_supported_decode_batch_sizes`` (e.g. Gemma4)
+        # may pad only to the nearest supported size >= num_reqs, so B=1 is not
+        # forced through a B=max graph.
+        #
+        # A model must declare only the buckets it has actually captured a decode
+        # trace for on this instance: padding to a bucket with no captured trace
+        # leaves the device in an undefined state. There is deliberately no
+        # separate "warmed" list to fall back from -- a bucket that is not warmed
+        # is not supported.
+        decode_buckets = getattr(
+            getattr(self, "model", None), "tt_supported_decode_batch_sizes", None
+        )
+        if decode_buckets:
+            decode_pad_to = next(
+                (
+                    int(b)
+                    for b in sorted(int(x) for x in decode_buckets)
+                    if int(b) >= num_reqs and int(b) <= self.tt_per_lane_max_num_seqs
+                ),
+                self.tt_per_lane_max_num_seqs,
+            )
+
         # Second dim of each block table is (ceil(max_model_len / block_size)).
         # Slice/pad to ``self.max_num_blocks_per_req``: slicing handles
         # over-wide tables when the total KV-cache limit is tighter than
--- a/tests/test_num_available_blocks.py
+++ b/tests/test_num_available_blocks.py
@@ -21,6 +21,11 @@ deterministic budget source:
   yields a fixed per-model budget, isolating this function from the
   per-SKU tables that live on the model class.
 
+``num_devices`` is now an explicit parameter (added by #425, "forward-compat
+prep for newer vLLM", which did not update these tests). Each call below passes
+``cfg.device_config.num_devices`` so a test's device count stays expressed in one
+place -- the fixture default of 1, or the explicit override a test sets.
+
 Sliding-window headroom is gated per model class by the
 ``_HYBRID_KV_CACHE_GROUPS_ENABLED`` attribute, read off the resolved model
 class. ``_model_budget`` sets it explicitly on the stub class (defaulting
@@ -91,7 +96,7 @@ def test_default_branch_no_sliding(cfg):
         patch("vllm_tt_plugin.worker.ttnn.get_arch_name", return_value="wormhole_b0"),
         _fallback_arch(),
     ):
-        n = get_num_available_blocks_tt(cfg)
+        n = get_num_available_blocks_tt(cfg, cfg.device_config.num_devices)
 
     # Default branch: max_tokens_all_users = 131072, plus block_size*batch
     # padding (64*32 = 2048). num_blocks = ceil(133120 / 64) = 2080.
@@ -118,7 +123,7 @@ def test_lane_mode_kv_shape_matches_per_lane_standard_dp(cfg):
         patch("vllm_tt_plugin.worker.ttnn.get_arch_name", return_value="wormhole_b0"),
         _fallback_arch(),
     ):
-        n = get_num_available_blocks_tt(cfg)
+        n = get_num_available_blocks_tt(cfg, cfg.device_config.num_devices)
 
     # Per-lane batch is 32 // 4 = 8.
     # Default tokens (131072) + batch padding (64 * 8 = 512) = 131584 tokens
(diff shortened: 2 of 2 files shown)


src/vllm_tt_plugin/gemma4_reasoning_fallback.py
sanjaradylov, line 1: I believe we need to follow the upstream [API](https://github.com/vllm-project/vllm/tree/main/vllm/reasoning), which looks well-structured: have the `src/vllm_tt_plugin/reasoning` subpackage and implement hooks here. Maybe even add deeper hierarchy:
```markdown
vllm_tt_plugin/
-> __init__.py
   ...
-> reasoning/
       -> __init__.py
       -> gemma/
           -> v4/
               -> __init__.py
               -> fallback.py
           -> # Other versions here
       -> # Other models here
```

Otherwise, there's a chance that the root is arbitrary in its structure, especially if we have more special cases like this in the future.

arginugaTT: Agreed, done — this now lives in `src/vllm_tt_plugin/reasoning/` with an `__init__.py` documenting what belongs there (wrappers over upstream parsers, never parsers of our own — #32 settled that).

I kept it flat as `reasoning/gemma4.py` rather than `reasoning/gemma/v4/fallback.py`, on the reasoning that upstream is itself flat (`vllm/reasoning/gemma4_engine_reasoning_parser.py`), and a three-level tree around a single 74-line module is structure ahead of demand. The moment there is a second wrapper — another model, or a v5 — the deeper hierarchy earns its keep and I would rather introduce it then, with two real cases to shape it. Happy to do it now if you would prefer the layout locked in up front.

Your underlying point about the root not becoming arbitrary drove a second change, though. Registration is now lazy, matching how upstream registers its own parsers:

```
ReasoningParserManager.register_lazy_module(
    "gemma4", "vllm_tt_plugin.reasoning.gemma4", "Gemma4ReasoningFallbackParser"
)
```

The previous version built the subclass eagerly in `register()`, which runs in every vLLM process for every model. Importing upstreams gemma4 parser measures **~1.7s**, so a Llama or Qwen deployment was paying that for a wrapper it would never use. Measured after the change: register costs **0.00ms** and the module is not in `sys.modules` until `--reasoning-parser gemma4` is actually selected.

Also worth recording why the wrapper exists at all, since it was fairly challenged on this PR. Both halves are verified against vLLM 0.25.1 rather than assumed:

1. Upstream parser, unterminated thought block → `content=None`, final answer stranded in `reasoning_content`. Terminated case is unaffected.
2. lm-eval scores that field alone by design — `parse_generations` does `content = message["content"] or ""`, and `ChatGeneration` subclasses `str` with *content* as the value specifically so the reasoning trace "never leaks into scoring".

So an empty `content` is scored as a wrong answer no matter how good the reasoning. The wrapper is a no-op whenever the model closes its thought block.

One known gap, called out in the module docstring: non-streaming only. `extract_reasoning_streaming` is unwrapped, so streaming clients see stock upstream behaviour.

sanjaradylov (dismissed):

viktorpusTT:
The change in parser was caused by dropping the custom parser in favor of upstream one. Upstream's parser behavior is almost certainly not a bug.

The proposed change is not a bugfix, but rather eval massaging, wearing a bugfix's clothes. Four tells:

* Fires only when content is empty, which is exactly and only the zero-score case.
* Keys on Final Answer: / Answer:. Those are conventions of the eval prompt, not of the model's output protocol. The protocol markers are <|channel> / <channel|>. Operating on protocol markers would be a parser fix; operating on prose formatting is answer extraction.
* Patches only the non-streaming path, which is the path evals use. A real parser defect would exist in both paths and would have to be fixed in both.
* The evidence offered is a score, not a protocol violation.

The eval should just not enable the parser. It's strictly less machinery than this PR. Verified the path:

* No --reasoning-parser → ParserManager.get_parser(None, None) returns None → serving sets parsers = [None] * num_choices → no splitting, full generation lands in content.
* Second effect pushing the same way: the gemma4 adapter's adjust_request is what sets skip_special_tokens = False. With no parser it never runs, so the default True holds and the detokenizer strips <|channel> itself.

arginugaTT:
You are right on all four tells — dropped the commit.

The third one I had already written into the docstring as a "known gap" without drawing the obvious conclusion from it: a real parser defect would exist in both paths. And keying on `Final Answer:` rather than `<|channel>`/`<channel|>` is answer extraction, not parsing.

Data agrees. Two runs on this branch with the wrapper removed:

- 31B, `r1_gpqa_diamond` ci-nightly, 40 samples: **0/40** empty content, 80/80 correct, PASS
- 12B, 10 samples: **90**, ratio 1.142 vs reference 78.8, PASS

PR is now two commits: the decode bucket pad and the `num_devices` test fix, both independent of this. No objection to the eval simply not enabling `--reasoning-parser`; it is strictly less machinery.

viktorpusTT (approved):
Thank you

PR #93, opened 2026-08-26
[bug] Release the model before closing the mesh in TTWorker.shutdown

### LINKED ISSUE #80: Worker shutdown closes the mesh while the model's device tensors are still alive
## Background

`TTWorker` owns a ttnn mesh device. The tt-metal model owns tensors, traces, and persistent captures allocated **on** that mesh. Releasing a device-resident object after its device has been closed is unsafe, so teardown must destroy the model first and close the mesh second.

`TTWorker.__del__` encodes exactly that and says so in a comment: it deletes `self.model_runner` first, then calls `close_mesh_device`.

## Problem

`TTWorker.shutdown()` (`src/vllm_tt_plugin/worker.py:488`) does it in the other order:

```python
runner = getattr(self, "model_runner", None)
if runner is not None:
    runner.shutdown()          # releases the persistent capture only
mesh_device = getattr(self, "mesh_device", None)
if mesh_device is not None:
    close_mesh_device(mesh_device, get_tt_config(self.vllm_config))
    self.mesh_device = None
```

`TTModelRunner.shutdown()` deliberately does **not** drop the model; its docstring says so. So when `shutdown()` returns, `self.model_runner` still holds the model and its device tensors, on a mesh that has just been closed. The actual release happens later, when `__del__` runs `del self.model_runner`, which is now *after* the close. That is the inversion the destructor's comment exists to prevent.

Two smaller leftovers in the same method: `self.device` and `self.device_config.device` still point at the closed mesh, since only `self.mesh_device` is cleared.

Choosing `shutdown()` as the hook is correct and should stay: upstream guarantees it on orderly teardown, and `__del__` is not guaranteed at interpreter exit.

## Suggested fix

In `shutdown()`: call `runner.shutdown()`, then drop the runner and model references (`self.model_runner = None`), clear `self.device` and `self.device_config.device`, and only then `close_mesh_device`. Keep the `self.mesh_device = None` idempotence guard.

## Acceptance

Extend the existing shutdown test with a stub runner whose teardown appends to the same order list as `close_mesh_device`, and assert the runner is torn down first. The current test uses a bare namespace with no model, so it cannot observe ordering.

## Related, worth fixing in the same change

The existing test does not pin *when* the close happens, only that it happens once across `shutdown()` plus `__del__`. Reverting `shutdown()` to its pre-fix body (mesh left to `__del__`) leaves the whole suite green. Assert the mesh is closed and `worker.mesh_device is None` after the first `shutdown()` and before `__del__` is invoked, then keep the existing idempotence assertion.

zhenghaoniTT:
## TL;DR

`TTWorker.shutdown()` closed the ttnn mesh while the model's device tensors were still alive. `TTModelRunner.shutdown()` now releases them explicitly first — the capture, the model, the KV caches, and the async-decode controller's back-reference — and the worker guarantees the close happens regardless.

> **Descoped after four review rounds.** Earlier revisions also drained in-flight async decode steps, called `gc.unfreeze()/gc.collect()`, cleared the async step queues, and had `__del__` delegate here. Review established that each either could not do what its comment claimed or introduced a new teardown failure. All four are removed; the residual gap is stated below rather than half-solved. Details in the last section.

## Details

The model owns tensors, traces and persistent captures allocated **on** the mesh, so teardown must release them before the mesh closes. `__del__` encodes exactly that and says so. `shutdown()` did the reverse: `runner.shutdown()` released only the persistent capture, then `close_mesh_device` ran with the model still live, and the real release happened later in `__del__`'s `del self.model_runner` — after the close.

Reordering a `del` cannot fix it. `TTModelRunner.__init__` builds `TTAsyncDecodeController(self)` (`model_runner.py:238`) and the controller keeps `self.runner` (`async_decode.py:172`), so refcounting never reclaims either object; whatever the attribute held is freed whenever the cyclic collector next runs — for interpreter teardown, after the devices closed.

### What it does

`TTModelRunner.shutdown()` is idempotent and releases, in a documented order,
the model and its caches while the mesh is open: the model-owned capture, then
`model` and `kv_caches`, then the async-decode controller's back-reference (so
an engine-held async output cannot walk back into a runner whose model is gone).
**The method's own comments are the description** — I am deliberately not
restating them here; across thirteen review rounds the recurring defect in this
PR was a mechanism described in three places, two of which then drifted.

`TTWorker.shutdown()` keeps no local reference to the runner and wraps the
release in `try/except` — the close is the step that must not be skipped,
because a mesh left open wedges the board's ethernet cores for the next process.
It also clears `self.device` and `device_config.device`: `init_device` sets both
alongside `self.mesh_device` (`worker.py:244-245`), so clearing only the last
left two live references to a closed device.

`shutdown()` stays the hook: upstream guarantees it on orderly teardown, and
`__del__` is not guaranteed at interpreter exit. `__del__` is unchanged, with a
comment on why it must *not* call `shutdown()`.

### What it deliberately does not do

It does not release an in-flight async decode submission. Those hold the read
destination and events of a device read that may still be running, and there is
no wait available here that can also be bounded against the close. They are left
to leak to process exit — the safe direction.

An earlier revision cleared `_pending_samples`, which is the opposite of this
tradeoff: between `execute_model` and `sample_tokens` the *runner* is the sole
holder of such a submission (the engine only takes it once `sample_tokens` hands
the wrapper over), and that window is exactly where a mid-step raise lands, via
`run_engine_core`'s `finally` around `run_busy_loop` (`vllm/v1/engine/core.py:1236-1242`). An unsampled step is pinned by **both**
`_pending_samples` and `_pending_async_steps` — `register_pending_async_step`
files it in the second — and neither is cleared now, so the submission survives
to process exit. The test pins that: reintroducing the clear fails it.

The cost is that the runner itself is not reclaimed promptly. That is not a
safety property — releasing the model before the close is, and it is asserted
directly.

### Tests

The old test used a bare namespace with no model, so it could only observe that
`shutdown()` was *called*. `test_worker_shutdown_releases_the_model_before_closing_the_mesh`
reads weakrefs **inside** the patched `close_mesh_device`, asking the question
directly — was the model gone before this call — rather than inferring it from
finalizer ordering.

Every liveness fact is **sampled at that same instant** rather than asserted
later: "was it released" stays true once true, but "is it still alive" does not —
after the runner drop the runner, its `_pending_samples` deque, the partial and
the submission form an isolated cycle that any gen-0 pass would collect. Cyclic
collection is disabled across the call too, so the window itself cannot be
perturbed.

Two review rounds were needed to get this right: the first restored `gc.disable()`
(reproduced by allocating past the gen-0 threshold inside the window — unguarded
fails, guarded passes), the second pointed out that two of the leak assertions
still ran *after* `gc.enable()`, where a crossing is in fact more likely because
CPython keeps counting allocations while collection is off. Verified by forcing a
collection in exactly that spot: the sampled form passes, the assert-after-enable
form fails.

`test_worker_shutdown_closes_the_mesh_even_if_the_runner_raises` pins the close.
Full mutation matrix:

```text
baseline                                   24 passed
self.model = None removed                   2 failed
capture release skipped                     3 failed
async_decode.runner cut removed             1 failed
_pending_samples.clear() reintroduced       3 failed
worker try/except removed                   1 failed
```

(`SimpleNamespace` is not weakref-able, hence the small `_Weakrefable` helper.)

### What review removed, and why

| Removed | Why |
|---|---|
| Bounded drain thread | On expiry the daemon thread keeps doing device work while the caller closes the same mesh, and dereferences state `shutdown` just cleared (`runner.model` → the misleading `AttributeError` at `async_decode.py:603`). |
| `gc.unfreeze()` + `gc.collect()` | Its stated purpose was disproved twice, and it can run tt-metal finalizers concurrently with the close. |
| Clearing the async step queues | The opposite of harmless: it frees the read destination and unsynchronized ttnn read events of a step the runner may still be the sole holder of, moments before the close. |
| `__del__` → `shutdown()` | `Thread.start()` at interpreter shutdown raises `RuntimeError` (`PythonFinalizationError` only on 3.13+; this repo targets `>=3.10,<3.14` and validates on 3.12), which escapes the destructor's `suppress(AttributeError)` and skips `close_mesh_device` entirely — a regression. |

## Related Artifacts

Resolves #80

## Validation

Host suite on a Blackhole quiet box (`bh_quietbox_2`), tt-metal `python_env` (Python 3.12, vLLM 0.25.1):

```text
$ ruff check src tests
All checks passed!

332 passed, 16 warnings in 6.33s
```

## Checklist

### DIFF
--- a/src/vllm_tt_plugin/model_runner.py
+++ b/src/vllm_tt_plugin/model_runner.py
@@ -75,6 +75,7 @@ import numpy as np
 
 logger = init_tt_logger(__name__)
 
+
 # Matches the upstream attention-layer naming convention used by registered
 # vLLM models (e.g. "model.language_model.layers.5.self_attn") as well as
 # bare "layers.5" forms used in TT spec hooks. The first capture group is
@@ -160,7 +161,7 @@ class TTModelRunner:
         self.device_config = vllm_config.device_config
         self._output_tokens_per_step = get_tt_output_tokens_per_step(vllm_config)
         self._is_block_output_model = is_tt_block_output_model(vllm_config)
-        self._persistent_capture_released = False
+        self._shutdown_complete = False
 
         if self.model_config.is_encoder_decoder:
             raise ValueError("Encoder-decoder models aren't yet supported for TT")
@@ -260,16 +261,35 @@ class TTModelRunner:
         )
 
     def shutdown(self) -> None:
-        """Deterministically release optional model-lifetime captures.
+        """Release what this runner holds on the mesh, while it is still open.
 
-        Called from ``TTWorker.shutdown`` while the mesh is still open. This
-        is intentionally not wired to ``__del__``: the runner sits in
-        reference cycles, so a destructor can fire during interpreter
-        teardown after the devices closed, where a device-side release is
-        worse than leaking the capture to process exit.
+        Called from ``TTWorker.shutdown`` because releasing a device-resident
+        object after its device closed is unsafe. Idempotent.
+
+        Explicit rather than a consequence of dropped references: the runner and
+        ``TTAsyncDecodeController`` reference each other, so refcounting never
+        reclaims either and a destructor runs whenever the cyclic collector gets
+        to it -- for interpreter teardown, after the devices closed.
+
+        Scope is the model and its caches. An in-flight async decode submission
+        is deliberately not released: its read destination and ttnn read events
+        may still be the target of a device read, and no wait for that can be
+        bounded against the close. Those leak to process exit, which is the safe
+        direction.
         """
-        if getattr(self, "_persistent_capture_released", False):
+        if getattr(self, "_shutdown_complete", False):
             return
+        self._shutdown_complete = True
+
+        # Neither ``_pending_samples`` nor ``_pending_async_steps`` is cleared: a
+        # step that ran but was not sampled sits in both, and between
+        # ``execute_model`` and ``sample_tokens`` this runner is its only holder
+        # -- the window a mid-step raise lands in -- so clearing them would free
+        # a buffer ttnn is writing into, moments before the close. The cost is
+        # that this runner is not reclaimed promptly, which is not a safety
+        # property; releasing the model below is.
+
+        # While the model is still alive, and before the mesh closes.
         release = getattr(
             getattr(self, "model", None), "release_persistent_capture", None
         )
@@ -278,7 +298,18 @@ class TTModelRunner:
                 release()
             except Exception:
                 logger.exception("Failed to release persistent model capture")
-        self._persistent_capture_released = True
+
+        self.model = None
+        self.kv_caches = []
+
+        async_decode = getattr(self, "async_decode", None)
+        if async_decode is not None:
+            # An engine-held async output keeps ``_controller``, so cut the
+            # controller's edge back to this runner: otherwise it walks into a
+            # runner whose model is gone and fails as a missing
+            # ``process_decode_output_host``.
+            async_decode.runner = None
+            self.async_decode = None
 
     @property
     def lane_batch(self) -> TTLaneInputBatch:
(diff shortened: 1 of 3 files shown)


viktorpusTT:
Shutdown frees the KV cache and the model while explicitly refusing to wait for the device read that is using them. Attempt the existing drain best-effort before releasing
anything:

```python
try:
    self.async_decode.wait_for_all_pending_async_steps()
except Exception:
    logger.exception("Async decode drain failed during shutdown")
```

optionally followed by a device synchronize, then release.

PR #94, opened 2026-08-26
[ref] Single-source the block-output sampling-neutralization table

### LINKED ISSUE #81: Block-output host-sampling neutralization omits logprobs, which can kill the engine
## Background

A block-output model must sample on device, because a host sampler produces one token and the contract needs a whole 256-token canvas. If any request in a batch carries a control that forces host sampling, the runner raises mid-step and the engine dies.

`TTScheduler` neutralizes those controls on requests that reached the engine without frontend validation. The table is at `src/vllm_tt_plugin/scheduler.py:114`:

```python
_BLOCK_HOST_ONLY_SAMPLING_NEUTRAL = (
    ("min_p", 0.0), ("min_tokens", 0), ("logit_bias", None),
    ("allowed_token_ids", None), ("bad_words", None), ("_bad_words_token_ids", None),
    ("presence_penalty", 0.0), ("frequency_penalty", 0.0), ("repetition_penalty", 1.0),
)
```

The comment above it asserts that any of these forces host sampling, which reads as a claim that the list is complete.

## Problem

It is not complete. `TTModelRunner.check_perform_device_sampling` (`src/vllm_tt_plugin/model_runner.py:1829`) also falls back to host sampling when logprobs are requested and the device count is outside a small set, and the documented DiffusionGemma configuration is `MESH_DEVICE=P150x4`, four devices.

So a repaired request carrying `logprobs=1` survives neutralization and then kills the engine, which is the exact failure the neutralization exists to prevent.

## The wider shape of it

The same predicate is currently written out in four places across three modules, in two different vocabularies (`SamplingParams` field names in the scheduler and in `validate_request`, batch state in the runner), plus a fifth copy, `TTPlatform.compat_sampling_required` (`src/vllm_tt_plugin/platform.py:1997`), which has no callers anywhere in the plugin or in vLLM 0.25.1. That dead copy is the one a maintainer is most likely to consult when extending the scheduler's table, because it is the shape-compatible one.

## Suggested fix

Make the runner's predicate the single definition and derive the rest from one exported table of `(field, neutral_value)`. Add `logprobs` and `prompt_logprobs` now. Delete `compat_sampling_required`, or comment it as retained for the fork and explicitly not the live predicate.

## Acceptance

A test asserting that every field the runner's predicate can trip on appears in the shared table, so the two cannot drift again.

zhenghaoniTT:
## TL;DR

`logprobs` was missing from the scheduler's neutralization table. It is **not** a live crash — the worker-side `disable_logprobs` guard already prevents it — so this is defence in depth. The substance is single-sourcing the table, deleting a dead and *differently wrong* fifth copy of the predicate, and replacing a list-versus-list drift test with a behavioral one that drives the real predicate.

> **Revised after review.** The first version claimed a fatal path and pinned it with a hand-built `SimpleNamespace` that bypassed the real code, plus a drift test that compared two hand-written lists. Both criticisms were correct; corrected below.

### What is actually true

`check_perform_device_sampling` falls back to host sampling whenever a logprob count is requested and the per-rank device count is outside `{8, 32}` — and the documented DiffusionGemma mesh is `MESH_DEVICE=P150x4`, four devices. A *requested* count is enough, including the `0` that OpenAI's `logprobs: true` produces, because the branch tests `max_lp is not None`.

### Why that is not a live crash

The worker-side guard runs first, and covers both `InputBatch` flavors unconditionally for block models:

```python
# model_runner.py:388, :400
disable_logprobs=self._is_block_output_model
# input_batch.py:421
if self.disable_logprobs:
    self.sampling.num_logprobs[req_index] = LOGPROBS_NONE_SENTINEL
```

so `max_num_logprobs` reports `None` and the predicate never sees a count. That is already covered by `test_worker_guard_neutralizes_logprobs_for_block_output` and `test_lane_batch_forwards_the_logprobs_guard`.

What the scheduler entry buys instead:

**Defence in depth**, and nothing more than that. This layer should not depend on the sentinel staying where it is, and the table should mirror the predicate so the two can be audited against each other — that is the whole justification.

> An earlier revision of this PR also claimed neutralizing here fixes a client-visible response shape (an empty logprobs array on a real completion). It does not, and review caught it: `LogprobsProcessor.from_new_request` snapshots `sampling_params.num_logprobs` into a plain int in the **frontend**, before `engine_core.add_request` (`v1/engine/llm_engine.py:274-276`, `async_llm.py:409-412`), and under multiprocess deployment the frontend holds a separately serialized `SamplingParams` anyway. A mutation inside EngineCore cannot reach it. Claim withdrawn from the code comments and from here.

### The dead fifth copy

The predicate was written out in four places across three modules in two vocabularies, plus `TTPlatform.compat_sampling_required` — **no callers anywhere** in the plugin or in vLLM 0.25.1, and a *different* logprobs rule:

```python
sampling_params.logprobs is not None and sampling_params.logprobs > 0
    and (num_devices == 1 or sampling_params.logprobs > 1)
```

It is the copy a maintainer extending the table would reach for, because it is the shape-compatible one, and copying it reproduces exactly this omission for `logprobs=0` and for four-device meshes. Deleted.

### Single source, split by consequence

The table moves to `config.py`, which every consumer already imports:

- `BLOCK_HOST_SAMPLING_FORCERS` — the mirror of the predicate. Every entry has a corresponding branch, so the two can be audited against each other.
- `BLOCK_PENALTY_NEUTRAL` — never fatal, and it does not change generation either. The cost is host waste that grows with the session: a non-neutral value flips `no_penalties`, and from then on every block decode step rebuilds the prompt and output token-history tensors in `_prepare_model_inputs` (`[rows, max_prompt_len]` and `[rows, max_output_len]`, only the second growing as generation does) — against a `max_model_len` of 262144 for the documented DiffusionGemma config — which are passed to the model and dropped there.

  > This rationale flipped twice, and the second flip was mine to own. In an earlier revision I replaced "built and discarded" with "they silently take effect", on the strength of `async_decode.py:521-524` passing `prompt_tokens`/`output_tokens` into `decode_forward`. I verified the call site and not the consumer. DiffusionGemma — the only model declaring `output_tokens_per_step > 1` — opens both `decode_forward` and `prefill_forward` with `del sampling_params` and never references `**kwargs` again, so the penalties cannot take effect. The original rationale was right, and it agreed with `platform.py:668-674` and `docs/diffusion-gemma.md:88-92`, which I had left contradicting my own new text. Restored.
- `BLOCK_UNSUPPORTED_RESPONSE_CONTROLS` — `prompt_logprobs`, which is neither. Review pointed out that an earlier revision put it in the forcers table under a comment saying it is *not* a forcer, which defeats the split: an entry there with no matching branch makes the next audit either delete it or distrust the header.
- `BLOCK_UNVALIDATED_SAMPLING_NEUTRAL` — the union, which is what the scheduler applies.

Conflating the fatal and the wasteful sets under one "host only" name is what let the fatal list read as complete when it was not.

### The drift pin is behavioral now

A list-versus-list assertion passes when a predicate branch is added and neither list is updated, which is fair criticism of the first version. Removed, along with the `HOST_SAMPLING_TRIGGER_FIELDS` frozenset it compared against — one hand-written list is better than two that agree with each other.

In its place, `test_repaired_bypassed_request_satisfies_the_device_sampling_predicate` builds a request carrying every field in the table, repairs it through the real `TTScheduler` method, adds it to a **real block `InputBatch`** (`disable_logprobs=True`, `output_tokens_per_step=16`), and asserts the **real** `check_perform_device_sampling` returns `False` before and `True` after. The request's fields are derived from the table, so a new entry extends the test; an entry with no value in the test fails a coverage assertion.

It runs against **both** batch shapes, which review showed is necessary. A block batch is the production one, but it is built with `disable_logprobs=True`, and that sentinel hides `logprobs` from the predicate entirely — so on that batch alone the two fields this table *gained* were invisible, and the assertions held with their neutralization removed. The second pass drops the guard so the logprobs branch is actually exercised. Verified by making the neutralization skip `logprobs` with the table intact: the test fails now, and did not before.

Also corrected in the code comments: the failure a host-sampling fallback produces is the explicit guard in `_get_output_tokens` ("Block-output step fell back to host sampling"), which raises before any sampling — not `_coerce_output_block`, which that guard means is never reached on this path.

`test_requested_logprobs_force_host_sampling_off_capable_meshes` is also rewritten to use a real non-block `InputBatch` — where the logprobs branch genuinely is the live path — instead of the stand-in that bypassed it.

## Related Artifacts

Resolves #81

## Validation

Host suite on a Blackhole quiet box (`bh_quietbox_2`), tt-metal `python_env` (Python 3.12, vLLM 0.25.1):

```text
$ ruff check src tests
All checks passed!

332 passed, 16 warnings in 6.34s
```

Both halves of the new pin confirmed to fail when broken:

```text
# A) neutralization skips min_p, table intact -> the behavioral half catches it
FAILED test_repaired_bypassed_request_satisfies_the_device_sampling_predicate

# B) add a table entry the test has no value for -> the coverage half catches it
AssertionError: update this test's non-neutral values for ['seed']
```

Experiment A is the one review made possible: the earlier revision built the request with `temperature=0.0`, and `SamplingParams` squashes `min_p` to 0.0 for a greedy request, so `min_p` was never actually exercised — removing its neutralization changed nothing, and the "drop min_p from the table" check I originally reported was really only tripping the coverage assertion. The test now builds with `temperature=1.0` and asserts every value survived construction.

## Checklist

### DIFF
--- a/src/vllm_tt_plugin/config.py
+++ b/src/vllm_tt_plugin/config.py
@@ -102,6 +102,56 @@ def is_tt_block_output_model(vllm_config: "VllmConfig") -> bool:
     return get_tt_output_tokens_per_step(vllm_config) > 1
 
 
+# ``SamplingParams`` fields that can force a step onto host sampling, and the
+# value that does not. Host sampling yields one token where a canvas is needed,
+# so ``_get_output_tokens`` raises and takes EngineCore with it.
+#
+# Mirrored from ``check_perform_device_sampling``; add a request-driven branch
+# there and you must add its field here. Pinned behaviorally, not by a second
+# list -- see ``tests/test_lane_input_batch.py``.
+BLOCK_HOST_SAMPLING_FORCERS: tuple[tuple[str, Any], ...] = (
+    ("min_p", 0.0),
+    ("min_tokens", 0),
+    ("logit_bias", None),
+    ("allowed_token_ids", None),
+    ("bad_words", None),
+    ("_bad_words_token_ids", None),
+    # Forces host sampling on any device count outside {8, 32} -- P150x4
+    # included -- and ``logprobs=0`` (OpenAI ``logprobs: true``) is enough. The
+    # worker-side ``disable_logprobs`` sentinel already blocks it; this entry is
+    # the second layer, so the table mirrors the predicate rather than depending
+    # on that sentinel.
+    ("logprobs", None),
+)
+
+# Response-shape controls a block-output model cannot honor. Kept out of
+# ``BLOCK_HOST_SAMPLING_FORCERS`` so that table stays an exact mirror of the
+# predicate.
+BLOCK_UNSUPPORTED_RESPONSE_CONTROLS: tuple[tuple[str, Any], ...] = (
+    ("prompt_logprobs", None),
+)
+
+# Penalties neither force host sampling nor change generation -- the model owns
+# its sampler and drops what it is handed. The cost is host waste: a non-neutral
+# value flips ``InputBatch.no_penalties``, and every block decode step then
+# rebuilds the prompt and output token-history tensors in
+# ``_prepare_model_inputs``, the output one growing with the session. The
+# frontend neutralizes them; mirror it here.
+BLOCK_PENALTY_NEUTRAL: tuple[tuple[str, Any], ...] = (
+    ("presence_penalty", 0.0),
+    ("frequency_penalty", 0.0),
+    ("repetition_penalty", 1.0),
+)
+
+# What TTScheduler applies to a request that reached the engine without frontend
+# validation.
+BLOCK_UNVALIDATED_SAMPLING_NEUTRAL: tuple[tuple[str, Any], ...] = (
+    BLOCK_HOST_SAMPLING_FORCERS
+    + BLOCK_PENALTY_NEUTRAL
+    + BLOCK_UNSUPPORTED_RESPONSE_CONTROLS
+)
+
+
 def store_tt_output_tokens_per_step(
     vllm_config: "VllmConfig", output_tokens_per_step: int
 ) -> None:
--- a/src/vllm_tt_plugin/model_runner.py
+++ b/src/vllm_tt_plugin/model_runner.py
@@ -1826,6 +1826,13 @@ class TTModelRunner:
         finish = self._pending_samples.popleft()
         return finish(grammar_output)
 
+    # Request-driven branches below decide host vs device sampling. A
+    # block-output model that lands on host sampling gets one token where the
+    # canvas contract needs ``_output_tokens_per_step``, so ``_get_output_tokens``
+    # raises. Two layers keep it away: the ``InputBatch`` is built with
+    # ``disable_logprobs`` for block models, and ``TTScheduler`` neutralizes
+    # ``config.BLOCK_HOST_SAMPLING_FORCERS`` on unvalidated requests -- extend
+    # that table when adding a request-driven branch here.
     def check_perform_device_sampling(
         self, is_decode: bool, has_structured_outputs: bool
     ) -> bool:
@@ -1867,6 +1874,8 @@ class TTModelRunner:
         # host sampling to compute full top-N from logits.
         max_lp = input_batch.max_num_logprobs
         if max_lp is not None:
+            # A *requested* count is enough, including the 0 that OpenAI's
+            # ``logprobs: true`` produces.
             if num_devices not in (8, 32):
                 return False
             if max_lp > 0 and not self.supports_topk_logprobs:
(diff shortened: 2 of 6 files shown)


viktorpusTT:
The PR does not single-source anything. The copy it moved had exactly one consumer in the same file. The two live duplicates named in issue #81  (platform._neutralize_model_owned_sampling and platform.validate_request) are
 untouched, so the number of hand-maintained copies of this knowledge is unchanged at four, and the penalty neutrals are now written in two places instead of one.
Drive _neutralize_model_owned_sampling from the tables. It is a mechanical for field, neutral in ... loop plus the ignored.append(f"{field}={value!r}") formatting the loop can produce, plus a fourth BLOCK_MODEL_OWNED_SAMPLING group for temperature / top_p / top_k / seed. Build validate_request's unsupported list from a rejection table the same way. Then the tables have real consumers and the title is earned. 

Issue #81's acceptance criterion is not met. The issue asks for "a test asserting that every field the runner's predicate can trip on appears in the shared  table, so the two cannot drift again." The PR's test asserts the opposite direction: every table field appears in the test. QA proved the gap empirically by inserting a new request-driven predicate branch and watching the full suite stay green at 332 passed.

Issue #89, opened 2026-08-26
Gemma4 registrations are dead no-ops and their comment is wrong for vLLM 0.25.1


viktorpusTT:
## Background

`_register_model_if_missing(ModelRegistry, arch, target)` registers an architecture name only when nothing already claims it, so it can never override an upstream registration.

The DiffusionGemma block in `register_tt_models` now states this correctly: upstream owns the bare name, do not pretend an if-missing registration overrides it. That reasoning is right.

## Problem

The Gemma4 block immediately above says the opposite. `src/vllm_tt_plugin/platform.py:1116` opens with:

> Gemma4 isn't in vLLM's upstream registry, so without an entry here [...]

and the loop at `:1140-1148` registers `Gemma4ForCausalLM`, `Gemma4ForConditionalGeneration`, and `Gemma4UnifiedForConditionalGeneration` to the TT target.

All three of those names **are** in upstream vLLM 0.25.1's registry (`vllm/model_executor/models/registry.py`, verified for each of the three). So all three `_register_model_if_missing` calls are no-ops, and upstream resolution does not find the TT class through them. The TT class is reached later, through the `TT`-prefix rewrite in the config hook.

Two consequences:

1. **The comment actively misleads.** A maintainer adding the next TT model reads both blocks, follows the older and more prescriptive one, registers a bare name upstream already owns, and then spends time working out why their adapter never loads.
2. **Gemma4 has the same early-config-hook exposure that #32 built a whole mechanism to solve for DiffusionGemma.** Upstream's `MODELS_CONFIG_MAP` contains all three Gemma4 names, and that lookup happens in `VllmConfig.__post_init__` well before `current_platform.check_and_update_config`. So upstream's Gemma4 config hook runs against a TT config, and `model_config._architecture`, `_model_info`, `runner_type`, and `convert_type` are all computed from upstream's Gemma4 class rather than the TT one.

## Suggested fix

- Rewrite the Gemma4 comment to describe 0.25.1 as it actually is, and either delete the three no-op calls or mark them explicitly as forward-compatibility no-ops.
- Then consider generalizing the DiffusionGemma mechanism rather than leaving it single-family: derive the `get_config` architecture rewrite map, the registration list, and the family membership test from one table of "architecture names TT claims". That closes the Gemma4 exposure with the mechanism that already exists, and makes adding the next model a one-line table edit instead of a four-site change.

PR #95, opened 2026-08-26
[doc] Correct the Gemma4 registration story and the stale constraint lists

### LINKED ISSUE #88: README constraints and registration convention are out of date after block-output support
## 1. A platform-wide rejection is documented as model-specific

`TTPlatform.validate_request` now rejects `prompt_embeds` for **every** TT model. The check sits before the block-output early return, so it applies to Llama, Qwen, Gemma, DeepSeek, GPT-OSS, everything. It did not exist before #32, where `prompt_embeds` failed later and deeper as a `NotImplementedError` from the input batch, so promoting it to a clean per-request rejection is a genuine improvement that nobody will find.

`README.md` "Operational Constraints" is the canonical list of what the platform refuses early, and it lists only the sibling check three lines above it in the same function: "Prompt logprobs are rejected at request validation time." (`README.md:463`). The only mention of `prompt_embeds` is inside `docs/diffusion-gemma.md`, where its placement in a DiffusionGemma paragraph implies it is a block-model restriction.

**Fix:** add a `prompt_embeds` bullet next to the prompt-logprobs bullet in README, and reword the DiffusionGemma sentence so the platform-wide and block-only rejections are distinguishable.

## 2. Two launch-time rejections are missing from the documented reject list

`docs/diffusion-gemma.md` enumerates what fails at launch ("Async scheduling, host sampling, custom logits processors, and non-uniproc executors are rejected at launch"). Two more exist and neither is listed:

- An explicit `--diffusion-config` is now a hard `ValueError` for block-output models. This is new in #32 and is a real vLLM 0.25.1 CLI flag, so an operator copying a launch line from upstream DiffusionGemma documentation gets a startup failure with no doc entry to check.
- `VLLM_USE_RUST_FRONTEND=1` is refused, because the Rust frontend runs HTTP outside Python and never invokes TT request validation.

Neither appears in `docs/diffusion-gemma.md` nor in README's constraints list.

## 3. The registration naming convention no longer holds

README states the plugin registers TT-prefixed architectures. After #32 DiffusionGemma breaks that in both directions: the plugin registers the bare `DiffusionGemmaForCausalLM`, and deliberately does **not** register the bare `DiffusionGemmaForBlockDiffusion` because upstream owns it.

README also lists only the two TT-prefixed DiffusionGemma names. Those are the names *after* the architecture rewrite, not what a real checkpoint's `config.json` declares, so an operator matching their checkpoint against the supported list finds no match for either real name.

**Fix:** add the checkpoint-side names to the DiffusionGemma entry with a pointer to how architecture resolution works, and note in the naming-convention paragraph that DiffusionGemma is the exception because its checkpoint architectures are rewritten before registry lookup.

zhenghaoniTT:
## TL;DR

The Gemma4 registration block claimed those architectures are absent from vLLM's registry and registered them on that basis. In 0.25.1 all three are upstream, so the calls were no-ops and the comment argued the opposite of the (correct) DiffusionGemma block below it. Drop the dead calls, state accurately what the `TT` prefix does and does not decide, and fix three stale documentation claims.

### Registration

`_register_model_if_missing` never overrides an existing registration, and vLLM 0.25.1 registers `Gemma4ForCausalLM`, `Gemma4ForConditionalGeneration` and `Gemma4UnifiedForConditionalGeneration` itself — verified against the installed wheel with `VLLM_PLUGINS=""` so the plugin's own registrations cannot mask an upstream absence. The three bare-name calls were therefore silent no-ops; the TT class is reached through the `TT`-prefix rewrite in `check_and_update_config`.

The comment mattered more than the dead code: it sits directly above the DiffusionGemma block, which states the opposite reasoning and is right.

### What the `TT` prefix actually buys

Narrower than the old comment implied. The rewrite decides which class the model loader instantiates. It does not touch what `ModelConfig.__post_init__` already derived from the *bare* architecture, and two of the three names are multimodal upstream (`Gemma4ForConditionalGeneration` → `gemma4_mm`, `Gemma4UnifiedForConditionalGeneration` → `gemma4_unified`, which subclasses it):

```text
Gemma4ForCausalLM                       supports_multimodal=False
Gemma4ForConditionalGeneration          supports_multimodal=True
Gemma4UnifiedForConditionalGeneration   supports_multimodal=True
```

So for a Conditional or Unified checkpoint `multimodal_config` is already built before the rewrite happens, `runner_type` and `convert_type` are upstream's, and upstream's `Gemma4Config` hook has already run against the TT config — all three names are in `MODELS_CONFIG_MAP`, and `VllmConfig.__post_init__` dispatches through it well before `check_and_update_config`.

That early-config exposure stays open. It is the same problem `_install_diffusion_gemma_architecture_patch` solves for one family; closing it for Gemma4 means generalizing that patch off a table of claimed architecture names, which is its own change.

### Documentation

- `validate_request` rejects `prompt_embeds` for **every** TT model — the check precedes the block-output early return (`platform.py:1963` vs `:1968`). README's constraint list had only the sibling prompt-logprobs bullet, and the sole mention of `prompt_embeds` sat inside a DiffusionGemma paragraph, implying a block-only restriction.
- The DiffusionGemma launch-reject list omitted an explicit `--diffusion-config` (a real vLLM 0.25.1 flag, so an operator copying an upstream launch line gets a startup failure with no doc entry) and `VLLM_USE_RUST_FRONTEND=1`.
- README listed only the post-rewrite `TT`-prefixed DiffusionGemma names, so an operator matching a real checkpoint's `config.json` against the supported list found no match for either name it declares.

## Related Artifacts

Resolves #88

Related to #89 — the dead calls and the wrong comment are fixed here; the Gemma4 early-config exposure the corrected comment documents is not.

## Validation

Host suite on a Blackhole quiet box, tt-metal `python_env` (Python 3.12, vLLM 0.25.1):

```text
$ ruff check src tests && ruff format --check src tests    # pinned 0.14.0
All checks passed! / 54 files already formatted

331 passed
```

Two new tests. `test_gemma4_bare_architectures_are_owned_by_upstream_vllm` asserts positive per-arch module identity (exact `module_name`, not a substring — `…models.gemma4` is a prefix of `gemma4_mm`, `gemma4_unified` and `gemma4_mtp`) and clears `TT_VLLM_BUILTIN_MODELS` / `TT_*_VER` first, since `vllm_tt_plugin.worker` runs the real registration at module import. `test_gemma4_conditional_archs_are_multimodal_upstream` pins the table above; it asserts registration unconditionally and skips the multimodal flag where the environment cannot import the multimodal processors, as on the host-only CI job.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,7 +406,18 @@ At startup the backend logs that it is running single-process lane-DP.
 ## Supported Model Families
 
 The plugin registers TT-prefixed model architectures backed by tt-metal model
-implementations. Current families:
+implementations. A checkpoint declares its own (bare) architecture name;
+`check_and_update_config` prepends `TT` before the registry lookup, which is
+why the names below are the registered ones rather than the ones in a
+`config.json`. The rewrite selects the class the model loader instantiates; it
+does not undo what `ModelConfig` already derived from the bare name upstream
+(architecture metadata, multimodal config, runner and convert types).
+
+DiffusionGemma is the exception in both directions: the plugin registers the
+bare `DiffusionGemmaForCausalLM`, and deliberately does *not* register the bare
+`DiffusionGemmaForBlockDiffusion`, because vLLM 0.25.1 owns that name and the
+plugin rewrites the architecture before registry lookup instead. Current
+families:
 
 - Llama 3.1 / 3.2 / 3.3 text models (`TTLlamaForCausalLM`)
 - Llama 3.2 vision models (`TTMllamaForConditionalGeneration`)
@@ -419,7 +430,10 @@ implementations. Current families:
   `TTGemma4ForConditionalGeneration`,
   `TTGemma4UnifiedForConditionalGeneration`)
 - DiffusionGemma block-output models (`TTDiffusionGemmaForBlockDiffusion`,
-  `TTDiffusionGemmaForCausalLM`)
+  `TTDiffusionGemmaForCausalLM`). A checkpoint's `config.json` declares the
+  bare `DiffusionGemmaForBlockDiffusion` or `DiffusionGemmaForCausalLM`; the
+  names here are what the rewrite above turns those into, so match your
+  checkpoint against the bare names.
 - DeepSeek V3 (`TTDeepseekV3ForCausalLM`)
 - GPT-OSS 20B / 120B (`TTGptOssForCausalLM`)
 
@@ -461,6 +475,9 @@ clear error before anything reaches the device:
 - Where chunked prefill is active, multimodal inputs are never split across a
   chunk boundary.
 - Prompt logprobs are rejected at request validation time.
+- `prompt_embeds` inputs are rejected at request validation time, for
+  every TT model. (Block-output models additionally reject
+  streaming-input sessions; see the DiffusionGemma document.)
 - Prefix caching is enabled only for models that declare TT support for it.
 - Async decode overlap is enabled only for models that declare the capability.
 - Multi-host MPI data parallelism is not supported.
--- a/docs/diffusion-gemma.md
+++ b/docs/diffusion-gemma.md
@@ -76,10 +76,20 @@ bucket has no servable execution shape and the failure is engine-fatal.
 
 Current support is synchronous, DP=1, one active sequence, on-device sampling
 for all model calls, and no scheduler chunked prefill; vLLM automatic prefix
-caching is disabled by the platform. Async scheduling, host sampling, custom
-logits processors, and non-uniproc executors are rejected at launch;
-`prompt_embeds` inputs and streaming-input (resumable) sessions are rejected
-per request; other unsupported per-request controls are listed below.
+caching is disabled by the platform.
+
+Rejected at launch: async scheduling, host sampling, custom logits processors,
+non-uniproc executors, an explicit `--diffusion-config` (the platform derives
+the canvas contract from the model's own capability declaration, so an operator
+copying a launch line from upstream DiffusionGemma documentation must drop that
+flag), and `VLLM_USE_RUST_FRONTEND=1` (that frontend runs HTTP outside Python
+and never calls TT request validation, so nothing would reject the requests
+listed below before they reach the engine).
+
+Rejected per request: streaming-input (resumable) sessions, for block-output
+models only. `prompt_embeds` inputs are rejected too, but that one is
+platform-wide -- every TT model refuses them, not just this one. Other
+unsupported per-request controls are listed below.
 
 HTTP sampling controls are accepted for OpenAI-client compatibility but ignored:
 
--- a/src/vllm_tt_plugin/platform.py
+++ b/src/vllm_tt_plugin/platform.py
@@ -1120,35 +1120,22 @@ def register_tt_models(register_test_models=False) -> None:
 
     # Gemma4 — text-only TT bridge.
     #
-    # Gemma4 isn't in vLLM's upstream registry, so without an entry here
-    # the upstream architecture resolver falls back to
-    # ``TransformersMultiModalForCausalLM`` (because ``hf_config !=
-    # hf_text_config`` for Gemma4's nested config — see
-    # ``ModelConfig._get_transformers_backend_cls``) and crashes on the
-    # ``_processor_factory`` assertion in the multimodal registry. The
-    # plugin's later ``TT``-prefix logic runs after that resolution, so
-    # it can't help.
+    # Only the ``TT``-prefixed aliases. Do not add the bare HF names: vLLM
+    # 0.25.1 registers all three itself and ``_register_model_if_missing``
+    # never overrides an existing entry, so those calls are no-ops.
     #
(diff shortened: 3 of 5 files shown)


viktorpusTT:
**Gemma 4 Conditional and Unified checkpoints accept image input, silently discard it, and answer on the leftover placeholder tokens.**
Chain of events for a 12B unified launch, all steps verified:
* vLLM builds ModelConfig. It looks up the bare name Gemma4UnifiedForConditionalGeneration, finds upstream's multimodal class, and so builds a multimodal_config (vllm/config/model.py:683, gated on _model_info.supports_multimodal). is_multimodal_model is a property over multimodal_config is not None
 (vllm/config/model.py:1616), so it is now True.
* Only then does TTPlatform.check_and_update_config run and rewrite the architecture to TTGemma4UnifiedForConditionalGeneration. Step 1 already happened and is not revisited.
* The OpenAI frontend now accepts image_url content parts, because multimodal_config exists. Upstream's Gemma 4 processor runs, converts the image to tensors, and injects image placeholder tokens into the prompt.

Fix: The right mechanism is reconcile, not replace, and it is general. Use llm.model_executor.models.interfaces.supports_multimodal(model_class). platform.py:1635 already resolves the
TT class after the prefix rewrite, and lines 1638 to 1665 already reconcile `supports_prefix_caching` and `supports_chunked_prefill` against upstream in exactly this shape. The check belongs there, as one more reconciliation:

```python
model_class, _ = get_model_architecture(vllm_config.model_config)   # 1635, exists

mm_config = vllm_config.model_config.multimodal_config
if mm_config is not None and not supports_multimodal(model_class):
    logger.warning(
        "%s declares a multimodal architecture, but the TT implementation "
        "(%s) is text only. Image and audio inputs will be refused; text "
        "inference is unaffected.",
        vllm_config.model_config.model,
        model_class.__name__,
    )
    vllm_config.model_config.multimodal_config = None
```

Restore the unified-checkpoint rationale to the comment, rewritten against the corrected mechanism: the 12B unified and the conditional checkpoints resolve to upstream's multimodal class, so `multimodal_config` is built and the frontend accepts images, while the TT bridge is text only. Note that `_CHUNKED_PREFILL_MODEL_TYPES` depends on this mapping.
Add the constraint to the README list this PR is already de-staling, beside the new `prompt_embeds` bullet at README.md:478.

README.md
```diff
@@ -419,7 +430,10 @@ implementations. Current families:
   `TTGemma4ForConditionalGeneration`,
   `TTGemma4UnifiedForConditionalGeneration`)
 - DiffusionGemma block-output models (`TTDiffusionGemmaForBlockDiffusion`,
-  `TTDiffusionGemmaForCausalLM`)
+  `TTDiffusionGemmaForCausalLM`). A checkpoint's `config.json` declares the
+  bare `DiffusionGemmaForBlockDiffusion` or `DiffusionGemmaForCausalLM`; the
+  names here are what the rewrite above turns those into, so match your
+  checkpoint against the bare names.
```
viktorpusTT, line 436: 1. **State the rule once, in the intro, in the reader's direction**, and end the
   intro with the colon that introduces the list. Replace
   [README.md:409-420](../README.md#L409-L420) with something of this shape:

   > The plugin registers TT-prefixed model architectures backed by tt-metal model
   > implementations. The names below are the registered ones: a checkpoint's
   > `config.json` declares the same name **without** the `TT` prefix, which the
   > plugin adds internally, so drop the prefix when matching your checkpoint
   > against this list. Current families:

2. **Fill in the four bullets that carry no names**, using the registered names
   from [platform.py:1085-1116](../src/vllm_tt_plugin/platform.py#L1085-L1116).
   Without this, edit 1 is still unusable for those families.

3. **Delete the duplicate rule from the DiffusionGemma bullet**
   ([README.md:433-436](../README.md#L433-L436)). Edit 1 covers it universally.
   The bullet goes back to just the two names.

4. **Delete the DiffusionGemma registration paragraph**
   ([README.md:416-420](../README.md#L416-L420)) from this section. It explains
   which bare names the plugin registers, which is maintainer knowledge with no
   operator consequence: the operator strips `TT` and matches either way. It is
   already recorded correctly, and more briefly, at
   [platform.py:1145-1151](../src/vllm_tt_plugin/platform.py#L1145-L1151). See
   also finding 5, which is about the reasoning inside this same paragraph being
   non-discriminating; deleting it resolves both.

5. **Add one line naming the launch error as the authoritative check**, for
   example: "If a checkpoint is not supported, startup fails with
   `No TT model architecture is registered for model: ...` and prints the full
   list of registered TT architectures."

viktorpusTT:
Both new test docstrings are the commit message, and they reference a comment that no longer exists. 

Fix: One line docstrings stating the invariant in the present tense ("Upstream vLLM owns the three bare Gemma4 architectures." / "The TT prefix does not change upstream's multimodal inspection of the bare name."), the mechanism as a short # comment matching the file's convention, and a path reference to platform.py:1122 instead of narrating "the comment". Leave the historical framing in the commit message. Annotate the fixture.

PR #104, opened 2026-08-31
fix: opt-in nulling of stale block-table row tails (rework of #103)

arginugaTT:
Rework of #103 per its closure rationale (unconditional 0-padding affected all models). The zeroing runs only for models declaring `tt_null_stale_block_table_tail = True`; all other models' tables are byte-for-byte unchanged **on every path**.

## Two commits

1. `9f1aec8` — opt-in nulling in `block_tables_for_rows`, gated at the model runner's per-group call on the model attribute. Root cause, fingerprint evidence, and before/after in the commit message; gemma4 declares the attribute in tt-metal `arg/gemma4_coherence_fix @ a570e36dfab`. Model-side masking was tried first and measured insufficient (21/32 — hybrid per-layer tables bypass the single-tensor kwarg). This is the only plugin change in the whole effort: `num_blocks_per_row` — the ground truth for a row's valid width — exists only in the plugin, so the masking cannot live anywhere else.
2. `b379451` — routes the same gate through **lane mode**: `slot_block_tables` forwards `null_stale_tail`, and both lane call sites (`_build_decode_input`, `_build_prefill_input`) gate on the same attribute. As first pushed, lane-mode deployments never executed the fix.

## Root cause (fingerprinted, WH T3K gemma-4-12B, 2 identical greedy prompts, isl 4k, chunk 2048)

vLLM's `move_row`/`condense` copies only `num_blocks` entries into a reused row; upstream GPU consumers mask by `num_blocks_per_row`, but the TT path materializes full width into device page tables, and gemma4's chunked-prefill continuation slices to full-prompt width — so a request condensed into a reused row consumes columns naming **another live request's blocks**, and its continuation KV writes clobber that request's cache (corrupt from decode step 2, concurrency >= 2, wherever the prefill budget makes prompts multi-chunk):

```
prefill A          row 0 -> [1..54]
prefill B chunk 1  row 1 -> [55..86]
condense() moves B to row 0 (copies 43 entries)
B chunk 2 consumed [55..97, 44,...,54]   <- tail = A's live blocks
vLLM at that step: num_blocks_per_row[0] = 43, consumed width = 54
```

## Evidence

**WH T3K (non-lane), serving path, 32 identical greedy prompts, 3 consecutive batches:**

| | before | after |
|---|---|---|
| correct | 11/32 | **32/32 x 3** |
| distinct outputs | 13-14 | **1** |
| degenerate | ~12/batch | **0** |

Also verified on 31B/T3K: 32/32 x 3 identical.

**BH LoudBox (lane mode), 12B, bounded prod shape, staggered >64Ki prompts:** with the lane amendment active, the gate fired on every call and nulled **9 genuinely nonzero stale tails** — the hazard is real on lane-mode Blackhole (latent above its 64Ki prefill budget) and the gated fix neutralizes it on both paths.

Why this presented as WH-only: BH specs run a 64Ki prefill budget so 4k prompts are single-chunk and the reused-row continuation path never executes there; WH's 2048 budget makes 4k prompts multi-chunk.

Host-only suite: **329 passed** at both commits.

### DIFF
--- a/src/vllm_tt_plugin/input_batch.py
+++ b/src/vllm_tt_plugin/input_batch.py
@@ -1116,15 +1116,23 @@ class TTLaneInputBatch(InputBatch):
     # ------------------------------------------------------------------
 
     def slot_block_tables(
-        self, rows: list[int], zero_gaps: bool, total: int, width: int
+        self,
+        rows: list[int],
+        zero_gaps: bool,
+        total: int,
+        width: int,
+        null_stale_tail: bool = False,
     ) -> list[torch.Tensor]:
         """Per-group block tables for ``rows`` (one row per slot), each padded
         to ``width`` (``max_num_blocks_per_req``). When ``zero_gaps`` is set,
         rows of ``range(total)`` not in ``rows`` are zeroed (empty decode slots
-        carry no blocks)."""
+        carry no blocks). ``null_stale_tail`` forwards to
+        ``block_tables_for_rows`` (lane mode builds both prefill and decode
+        tables through here, so the opt-in must ride along or lane-mode
+        deployments never execute the fix)."""
         occupied = set(rows)
         sel = list(range(total)) if zero_gaps else rows
-        out = self.block_tables_for_rows(sel, width)
+        out = self.block_tables_for_rows(sel, width, null_stale_tail=null_stale_tail)
         if zero_gaps and len(occupied) < total:
             gap = torch.ones(total, dtype=torch.bool)
             gap[list(occupied)] = False
@@ -1206,7 +1214,15 @@ class TTLaneInputBatch(InputBatch):
         input_tokens = torch.from_numpy(tokens_np)
 
         block_tables_per_group = lane_batch.slot_block_tables(
-            occupied, zero_gaps=True, total=total, width=runner.max_num_blocks_per_req
+            occupied,
+            zero_gaps=True,
+            total=total,
+            width=runner.max_num_blocks_per_req,
+            null_stale_tail=bool(
+                getattr(
+                    getattr(runner, "model", None), "tt_null_stale_block_table_tail", False
+                )
+            ),
         )
         rows_all = list(range(total))
         tt_sampling_params = lane_batch.slot_sampling_params(rows_all)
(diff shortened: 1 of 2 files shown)


viktorpusTT:
TT does not inherently need to access leftover block-table columns. The reported Gemma 4 continuation path explicitly contains logic intended to prevent that. The current change is containment, not a demonstrated root-cause fix.

I fear there is an underlying issue and by doing this neutralization, we just keep it there. Zeroing the page table converts cross-request access into shared scratch-block access. It does not eliminate the invalid access. We know exactly how the loaded gun is created and exported, but we still cannot point at the thing that pulls its trigger.

I suggest the following debug:

For one deterministic two-request run with different prompts:

1. Log request ID, row, valid count, and full block table before and after remove, append, and move.
2. Track physical block ID -> owning request ID.
3. Immediately before fill, log:
   * eager versus traced path
   * hybrid and bounded flags
   * actual block size
   * real valid_seq_len
   * K/V input row count
   * full page table
   * capped page table
   * exact chunk_page_table
4. Synchronize and fingerprint K and V for:
   * B's intended blocks 32 through 42
   * every stale physical block referenced by columns 43 through 53
   * A's corresponding live blocks
5. Capture those fingerprints immediately before and after each individual B fill.

The result becomes unambiguous:

* A-owned KV changes during B's fill: wrong writer proven.
* A-owned KV does not change, but valid attention changes when the stale tail changes: wrong reader proven.
* Neither happens: investigate persistent page-table copying, trace replay, or another model-state path.

arginugaTT:
Closing unmerged after root-cause work landed model-side. A/B verification on the gated gemma4 serving path showed stock main byte-identical to this branch everywhere tested (identity smoke, conc3 distinct 12k, b32@4k 32/32 byte-equal, 128k coherent, at PR-baseline perf), and the corruption this nulling contained was root-caused to two gemma4 model-layer bugs (per-layer page-table slot ordering + sliding-tail trace clobber), both fixed on tt-metal arg/gemma4_coherence_fix. The gemma4-side opt-in attribute was reverted (4ad79b54f0e), so this change protects nothing on the current stack. Review critique (nulling = containment, not root cause) was correct.

PR #99, opened 2026-08-27
[ci,doc] Add tools to automate vLLM-version comparison & adaptation.

sanjaradylov:
## TL;DR
Add a reusable, on-demand framework for assessing (and applying) vLLM version bumps: a phase-by-phase playbook plus a scheduled release watcher that files a tracking issue when upstream ships a newer release.

## Details
The plugin binds vLLM's internal v1 APIs, so every minor bump can break it and each one needs a compatibility pass. This PR captures that pass as repo-native tooling:

- **`.github/prompts/vllm-upgrade-assessment.prompt.md`** — an on-demand playbook (not auto-loaded, so zero cost to unrelated sessions). Phases:
  → pin env
  → map the coupling surface (grep re-derivation + a snapshot cross-check)
  → diff the coupled files (installed baseline vs `refs/tags/<TARGET>`)
  → classify by severity
  → synthesize the change plan
  → validate (ruff → build via `docs/install-vllm-tt.sh` → host-only pytest)
  → record & ship.

  Kickoff is a one-liner:
  > Follow .github/prompts/vllm-upgrade-assessment.prompt.md; assess this plugin against vLLM vX.Y.Z.

  It is also `/`-invocable in Copilot for VS Code. Works with Claude or Copilot; `tt-buddy` skills are an optional accelerant, not a dependency.

- **`.github/workflows/vllm-release-watch.yml`** (**in progress, needs an approval from reviewers**) — weekly (+ manual) job that compares the `vllm==` pin in `docs/install-vllm-tt.sh` to vLLM's latest stable release and, if newer, opens a **deduped** tracking issue pointing at the playbook with TARGET/BASELINE filled in. Uses `GITHUB_TOKEN` + `issues: write`; no secrets. The issue body is identical to the playbook's kickoff line, so manual and automated entry points converge.

## Related Artifacts
Relates to #64 (the framework's first target — vLLM 0.26.0). The 0.26.0 upgrade itself ships separately.

## Validation

The framework was exercised end-to-end two sessions: 0.25.1→0.26.0 #96 (1 fix; 329 host-only tests pass on a real 0.26.0 build) and 0.26.0→0.28.0 (2 scheduler breakages identified).

## Checklist

- [x] This PR is focused on a single logical change (the upgrade-assessment framework).
- [x] ~I added or updated tests where applicable~ — N/A (docs + CI workflow; YAML + shell logic validated above).
- [x] I ran the relevant checks and recorded them above.
- [x] I updated documentation where applicable (the playbook is the doc).

PR #105, opened 2026-09-01
[doc] Add AGENTS.md and pin pytest below 9 in the dev extra

viktorpusTT:
## TL;DR

Adds `AGENTS.md`, the agent-facing contribution contract this repository did
not have, and pins `pytest>=8,<9` in the `dev` extra so the pin AGENTS.md cites
is real.

## Details

This repository is the primary home for TT vLLM plugin work but carried no
`AGENTS.md`, `CLAUDE.md` or `.cursor` configuration, while `tt-metal` has
`.github/copilot-instructions.md` plus 16 path-scoped instruction files. Agents
working here therefore defaulted to two wrong behaviours that the review history
shows repeatedly:

- mirroring changes into `tenstorrent/vllm`, which was deprecated on 2026-08-24
  (tenstorrent/vllm#476) and whose in-tree plugin copy is no longer a reference
- gating behaviour on model-type allowlists rather than the `model_capabilities`
  dictionary the platform actually resolves

`AGENTS.md` covers: the plugin's entry-point surface and the rule that TT
features do not land in vLLM core; the two environments and why
`docs/install-vllm-tt.sh` must be sourced and owns the dependency set; the
`model_capabilities` gating contract with each key's default and effect, plus
the leftover identity gates that exist but must not be copied; the recurring
review themes; and the `TTPlatform` class-state and conftest traps that leak
between tests.

The `pyproject.toml` change is one line plus a comment. `pytest 9` `caplog`
attaches its own capture handler to non-propagating loggers and double-captures
records, which breaks `tests/test_logger.py`. CI already installs `pytest<9`
explicitly; the `dev` extra did not, so `uv pip install -e ".[dev]"` could float
to 9 and produce failures CI never sees. AGENTS.md states that pin, so the pin
has to exist.

Deliberately excluded, to keep this to one logical change:

- `ci.yaml` has a dead `push: branches: [dev]` trigger (default branch is `main`,
  no `dev` branch exists on the remote). Documented in AGENTS.md section 5, not
  fixed here.
- Markdown SPDX headers. No `.md` in this repository carries one, so AGENTS.md
  documents the Python convention only rather than introducing a new one.

## Related Artifacts

Not duplicating an existing PR: no open or merged PR adds `AGENTS.md`
(`gh pr list --search "AGENTS"` returns only unrelated matches). It does not
overlap #99, which added `.github/prompts/vllm-upgrade-assessment.prompt.md`
and `.github/workflows/vllm-release-watch.yml` for version-bump assessment;
AGENTS.md does not restate that tooling. It does not conflict with #101
(vLLM 0.28.0): AGENTS.md names no specific vLLM version, so it does not go
stale when that lands.

## Validation

Documentation and packaging metadata only. No source or test behaviour changes.

Every concrete claim in AGENTS.md was checked against the tree at
`origin/main`: the two entry points in `pyproject.toml`; the seven `_install_*`
and `_pin_v1_model_runner` helpers in `platform.py`; the five
`model_capabilities` keys and their `.get` defaults; the five attribute names in
`_TT_PLATFORM_CONFIG_ATTRS`; the `tests/tt/conftest.py` fixture neutralization;
the CI job matrix; and `vllm-model-tests.yaml` in tt-metal being
`workflow_dispatch` with `vllm-tt-plugin-ref` defaulting to `main`.

```text
$ pre-commit run
check toml...............................................................Passed
ruff check...........................................(no files to check)Skipped
ruff format..........................................(no files to check)Skipped
```

## Checklist

- [x] I added or updated tests where applicable. Not applicable: docs and
      packaging metadata only.
- [x] I ran the relevant checks such as unit tests and Actions and recorded
      them above.

---

AI assistance was used to draft `AGENTS.md` and to verify its claims against the
tree. Every line was reviewed by the submitter.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -1,0 +1,411 @@
+# Agent Instructions for `vllm-tt-plugin`
+
+These instructions apply to all AI-assisted contributions to
+`tenstorrent/vllm-tt-plugin`. Read this file before changing anything in the
+repository.
+
+## 1. What this repository is
+
+`vllm-tt-plugin` integrates Tenstorrent hardware into vLLM through vLLM's
+standard out-of-tree plugin mechanism. Two entry points, both declared in
+`pyproject.toml`, are the whole hook surface:
+
+- `vllm.platform_plugins:tt` maps to `vllm_tt_plugin.entrypoints:platform_plugin`
+- `vllm.general_plugins:tt_model_registry` maps to `vllm_tt_plugin.entrypoints:register`
+
+Both gate on `import ttnn`. When `ttnn` is not importable, the plugin declines
+to activate and vLLM proceeds as if it were not installed.
+
+The plugin is self-contained by design. Model registration, platform detection,
+config validation, scheduling, worker execution, model loading, async decode,
+and all three execution modes live here:
+
+- single-process non-DP
+- single-process lane-DP
+- standard multi-process DP (per-rank mesh and `TT_VISIBLE_DEVICES` discovery)
+
+See `docs/SCHEDULING.md` and `src/vllm_tt_plugin/utils/dp_discovery.py`.
+`src/vllm_tt_plugin/launcher.py` is retained MPI / tt-run and is not hooked by
+vLLM; it is not a fourth live mode.
+
+| Term | Meaning |
+|---|---|
+| DP | Data parallelism |
+| standard DP | Multi-process, one device mesh per rank |
+| lane-DP | Single-process, multiple TT lanes in one engine |
+| mesh | Device mesh (`MESH_DEVICE`), not a service mesh |
+| `python_env` | tt-metal virtualenv at `$PYTHON_ENV_DIR` |
+
+Do not open TT-feature pull requests against `vllm-project/vllm` or the
+deprecated `tenstorrent/vllm` fork. When plugin entry points cannot express
+the behaviour, runtime-monkeypatch vLLM from this repo next to the existing
+`_install_*` helpers in `src/vllm_tt_plugin/platform.py`. An upstream-general
+bug may still be fixed in vLLM core (see the TODO on
+`_install_tt_harmony_truncation_patch`). Neighbors for a new patch:
+`_install_diffusion_gemma_architecture_patch`,
+`_install_tt_harmony_truncation_patch`, `_pin_v1_model_runner`, and the
+`_install_block_output_*` functions.
+
+### Repository boundaries, and what changed recently
+
+Three repositories used to carry TT vLLM work. Their roles now:
+
+| Repository | Role today |
+|---|---|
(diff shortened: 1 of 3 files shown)


AGENTS.md
```diff
@@ -0,0 +1,411 @@
...
+`numpy<2` wins because tt-metal fixes it and it cannot be moved from here.
+
+**When bumping the pinned vLLM version, all of the following must change
+together:**
+
+- `docs/install-vllm-tt.sh`, the `common.txt` URL (line with `v0.x`)
+- `docs/install-vllm-tt.sh`, the `vllm==0.x` install line
+- `README.md`, the stated version
+- `docs/vllm-overrides.txt`, re-checked against the new `common.txt` (upstream
+  may have relaxed the opencv floor)
```
sanjaradylov, line 170: - Missing `docs/diffusion-gemma.md` (the two "pinned vLLM 0.x" / "parsers shipped by vLLM 0.x" lines).
- `README.md` has two references (the stated version and the launcher.py comment).
- Let's also require verification after installation to prevent env drift:
  `python -c "import vllm; print(vllm.__version__)"`

sanjaradylov (approved):
I can see that my Copilot CLI can automatically detect AGENTS.md at session startup. Not sure about Claude and friends.

AGENTS.md
viktorpusTT: All three applied in 341c4fd.

`docs/diffusion-gemma.md` (pinned-build line and the `gemma4` parsers line) and the second `README.md` reference (the `launcher.py` line in the package layout) were both missing. The post-install check is in, with a note on why it is worth doing: the pin is spread across a sourced script that installs with `--no-deps`, so what resolved and what the docs claim can diverge silently. I left the `compat/vllm-X.Y.Z` guidance in the README alone, since it is generic and survives a bump.

Verifying your point turned up one more. `install-vllm-tt.sh` has `torchvision==0.26.0` a few lines above `vllm==0.26.0`. Same string, unrelated pins: torchvision tracks tt-metal `requirements-dev.txt`. A find-and-replace on the version quietly breaks the torch pair, so the checklist now says explicitly not to touch it.

On the tooling question: Claude Code does not read `AGENTS.md`. Its docs are explicit ("Claude Code reads CLAUDE.md, not AGENTS.md"), and it discovers `CLAUDE.md`, `.claude/CLAUDE.md`, `CLAUDE.local.md` and `.claude/rules/*.md`. The supported shim is a `CLAUDE.md` that does `@AGENTS.md`, which is what `tenstorrent/vllm` already does. Added that file, plus a line in section 11 saying instructions go in `AGENTS.md` and never in `CLAUDE.md`, so the two toolchains cannot drift.

As per Google (yes, I'm that old), current Cursor honours `AGENTS.md`.

PR #106, opened 2026-09-02
fix google/gemma-4-E2B-it AmbiguousGlobalPerLayerAttributeError

yamt:
transformers 5.15.0 raises AmbiguousGlobalPerLayerAttributeError on head_dim access for heterogeneous configs like gemma-4.

> transformers.integrations.heterogeneity.configuration_utils.
> AmbiguousGlobalPerLayerAttributeError: 'head_dim' is a per-layer
> attribute and may vary across layers. Access it via the individual
> layer configs instead (e.g. config.per_layer_config[i].head_dim).
> To read the global config value from config.head_dim anyway, set
> `allow_global_per_layer_attribute_access` to `True` on the config.
> Warning: only do this if the caller can safely handle heterogeneous
> configs; code that assumes a homogeneous model may use the global
> value incorrectly.

## Validation

```text
<command and result>
```

## Checklist

- [ ] I added or updated tests where applicable.
- [ ] I ran the relevant checks such as unit tests and Actions and recorded them above.

### DIFF
--- a/docs/vllm-overrides.txt
+++ b/docs/vllm-overrides.txt
@@ -16,3 +16,6 @@
 # (https://github.com/tenstorrent/tt-metal/issues/30193).
 opencv-python-headless==4.11.0.86
 numpy>=1.24.4,<2
+
+# cf. https://github.com/vllm-project/vllm/pull/49797 (vllm v0.28.0 has this)
+transformers>=5.5.3,<5.15.0
--- a/docs/vllm-overrides.txt
+++ b/docs/vllm-overrides.txt
@@ -17,5 +17,17 @@
 opencv-python-headless==4.11.0.86
 numpy>=1.24.4,<2
 
-# cf. https://github.com/vllm-project/vllm/pull/49797 (vllm v0.28.0 has this)
+# vLLM 0.26.0 reads Gemma 4's head_dim off the global HF config
+# (transformers_utils/model_arch_config_convertor.py get_head_size, and
+# model_executor/models/gemma4.py). transformers 5.15.0 made head_dim a per-layer
+# attribute on heterogeneous configs and raises AmbiguousGlobalPerLayerAttributeError
+# on that read, so google/gemma-4-* fails during ModelConfig construction, before any
+# TT code runs. vLLM 0.28.0 carries the fix (vllm-project/vllm#49797).
+#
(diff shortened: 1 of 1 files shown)


docs/vllm-overrides.txt
```diff
@@ -16,3 +16,6 @@
 # (https://github.com/tenstorrent/tt-metal/issues/30193).
 opencv-python-headless==4.11.0.86
 numpy>=1.24.4,<2
+
+# cf. https://github.com/vllm-project/vllm/pull/49797 (vllm v0.28.0 has this)
```
viktorpusTT:
```suggestion
# vLLM 0.26.0 reads Gemma 4's head_dim off the global HF config
# (transformers_utils/model_arch_config_convertor.py get_head_size, and
# model_executor/models/gemma4.py). transformers 5.15.0 made head_dim a per-layer
# attribute on heterogeneous configs and raises AmbiguousGlobalPerLayerAttributeError
# on that read, so google/gemma-4-* fails during ModelConfig construction, before any
# TT code runs. vLLM 0.28.0 carries the fix (vllm-project/vllm#49797).
#
# Retarget this ceiling when the pinned vLLM moves; do not just delete it. common.txt
# declares no upper bound, so an uncapped install resolves the newest transformers
# release, which is validated against neither the pinned vLLM nor the tt-metal env
# (tt_metal/python_env/requirements-dev.txt pins transformers 5.12.1). The floor
# restates vLLM's own: --override replaces a requirement instead of intersecting it.
transformers>=5.5.3,<5.15.0
```

viktorpusTT:
The fix is correct and I verified it independently: `uv pip compile` against `common.txt`
from the `v0.26.0` tag resolves `transformers==5.16.1` on `main` today and `5.14.1` with
this override. `AmbiguousGlobalPerLayerAttributeError` subclasses `RuntimeError`, not
`AttributeError`, so `getattr(self.hf_text_config, "head_dim", None)` in
`vllm/transformers_utils/model_arch_config_convertor.py` does not swallow it and every
`google/gemma-4-*` launch dies during `ModelConfig` construction.

viktorpusTT (approved):
Thank you. One suggested comment update and we're ready to merge.

yamt:
> Thank you. One suggested comment update and we're ready to merge.

i hit the apply button. thank you.

PR #111, opened 2026-09-05
[bug] Reject non-greedy requests for greedy-only TT samplers

nkapreTT:
## TL;DR

Reject non-greedy requests at the API validation boundary when a TT model explicitly declares its on-device sampler greedy-only. Existing model behavior is unchanged when the new capability is absent.

## Details

Generated Quetzal performs device-side argmax and fails closed if temperature, top-p, or bounded top-k sampling reaches the model runner. Today that exception occurs in EngineCore and tears down the server. This change consumes the semantic `supports_non_greedy_sampling_on_device` model capability, stores the resolved value on TTPlatform, and raises before engine dispatch. The capability defaults to true only to preserve the established contract of already-enrolled device samplers; restricted models must opt out explicitly. Invalid non-boolean declarations fail startup.

AI assistance was used. The submitter must review and defend every changed line before merge.

## Related Artifacts

Paired Quetzal capability declaration: `tenstorrent/tt-quetzalcoatlus:nkapre/declare-greedy-only-device-sampling` (PR link will be added once created).

Observed incident: GPT-OSS-120B generated Quetzal returned HTTP 500 and EngineCore exited after a canonical AIME request used temperature=1.0.

## Validation

```text
PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest -q tests/test_block_request_validation.py
72 passed

PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest -q tests/ --ignore=tests/tt
371 passed, 14 warnings

.venv/bin/pre-commit run
PASS (all hooks)
```

Device/API validation will be attached after the paired Quetzal capability is staged on GPT-OSS-120B, P300X2.

## Checklist

- [x] I ran the relevant checks and recorded them above.
- [x] Documentation is not required; this is an internal model capability consumed at startup.

### DIFF
--- a/src/vllm_tt_plugin/platform.py
+++ b/src/vllm_tt_plugin/platform.py
@@ -1243,6 +1243,7 @@ class TTPlatform(Platform):
     _standard_dp_visible_device_groups: ClassVar[list[str] | None] = None
     _standard_dp_mesh_grids: ClassVar[dict[str, tuple[int, int]]] = {}
     sample_on_device_mode: ClassVar[Literal["all", "decode_only"] | None] = None
+    non_greedy_decoding_on_device: ClassVar[bool] = True
     # Stored as a weakref in production so a torn-down engine's config stops
     # tripping the one-engine-per-process guard once nothing else holds it;
     # tests may seed a direct config object.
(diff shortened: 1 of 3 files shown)


nkapreTT:
Paired capability declaration is now https://github.com/tenstorrent/tt-quetzalcoatlus/pull/64 (stacked on its GPT compiler dependency).

viktorpusTT:
Let me offer an alternative approach: Fallback to host sampling: https://github.com/tenstorrent/vllm-tt-plugin/pull/114
There already is a fallback mechanism in the plugin, so we're not introducing a new mechanism.

viktorpusTT:
@nkapreTT What do you think about the alternative approach in #114?

nkapreTT:
Closing in favour of #114.

#114's two objections to this approach are both correct, and I verified them against `origin/main`:

1. `TTPlatform.validate_request` only runs under the Python HTTP frontend. The Rust-frontend guard sits inside `if is_block_output_model:` (`platform.py:1644`), and generated Quetzal is non-block-output (`output_tokens_per_step` absent => 1), so under `VLLM_USE_RUST_FRONTEND=1` this check would be dead code for exactly the model shape that motivated it.
2. Reading a mutable `TTPlatform` class attribute per request breaks this repo's own stated invariant that "AR engines may share a process freely". #114's per-instance snapshot at `TTModelRunner.__init__` is the better pattern.

Enforcing in `check_perform_device_sampling` is the right location — it already routes eight other unsupported-feature cases to host sampling and runs unconditionally per step.

One thing worth carrying forward into #114: the shared boolean capability cannot express the real constraint. Generated Quetzal supports temperature, full-vocabulary nucleus, top-k <=64 and seeds on device, and raises on `min_p`, the three penalties, logprobs, and `top_k > 64`. So `temperature=0, presence_penalty=0.5` is a *greedy* request that passes either PR's check and still terminates EngineCore. A declared envelope would close that; a boolean will not.

PR #114, opened 2026-09-07
[bug] Fall back to host sampling for greedy-only device samplers

viktorpusTT:
## TL;DR

Route a non-greedy request against a greedy-only on-device sampler to vLLM's existing per-step host-sampling fallback, instead of rejecting the request at the API boundary. Alternative to #111, same underlying capability, different enforcement point; see Details for why.

## Details

**Problem.** Some TT models perform sampling on the Tenstorrent device rather than the host. A device sampler that only implements greedy decoding (argmax) cannot execute a `temperature`/`top_p`/`top_k` request. Before either this PR or #111, such a request reached the device unchecked and crashed `EngineCore` (observed incident, per #111: GPT-OSS-120B generated Quetzal, HTTP 500, engine exit, on a `temperature=1.0` request).

**Why not #111's approach.** #111 fixes the same incident by adding a `model_capabilities` key, `supports_non_greedy_sampling_on_device`, and rejecting the offending request in `TTPlatform.validate_request`. Reviewing that approach surfaced two defects specific to enforcing it there:

- `TTPlatform.validate_request` only runs through vLLM's Python HTTP frontend. An existing comment in `platform.py` states this: "The Rust frontend handles HTTP outside Python, so nothing runs `TTPlatform.validate_request`." For a non-block-output model (the shape of the model in the reported incident) with `VLLM_USE_RUST_FRONTEND=1`, the new check in #111 never runs at all, and the original crash can still happen.
- The resolved capability is read live off a process-global `TTPlatform` class attribute on every request. Two AR engines sharing one process (this plugin's own comment: "AR engines may share a process freely") can stomp on each other's declared capability.

**This PR's approach.** Same `model_capabilities` key (`supports_non_greedy_sampling_on_device`, default `True` for backward compatibility), same resolution in `TTPlatform._apply_check_and_update_config`. The difference is where it's enforced: `TTModelRunner.check_perform_device_sampling` (`model_runner.py`) already decides, per step, whether a batch can use on-device sampling or must fall back to host sampling, for several other unsupported features (`bad_words`, `min_p`, `logit_bias`, `allowed_token_ids`, `min_tokens`, custom logits processors, structured outputs, unsupported logprobs shapes). This PR adds one more condition there: if the model is greedy-only and any active row in the batch is non-greedy (`InputBatch.all_greedy`, already defined, previously unused), fall back to host sampling for that step, exactly like every other case already handled by that method.

Side effects of moving the check there, both verified by reading the code, not asserted:

- `check_perform_device_sampling` runs in the model runner's per-step execution path unconditionally, so this works under the Rust frontend too, not just the Python frontend.
- `TTModelRunner.__init__` snapshots the capability once, into an instance attribute, at worker construction (`self.supports_non_greedy_sampling_on_device = getattr(TTPlatform, ..., True)`), the same pattern already used for `sample_on_device_mode`. A later-starting sibling engine's class-attribute mutation does not retroactively change an already-constructed instance's snapshot.

**Trade-off, stated plainly.** `check_perform_device_sampling` returns one bool per step's batch, not per request, matching its existing contract for every other host-only condition. If one row in a step's batch is non-greedy against a greedy-only model, the whole batch (including greedy rows sharing that step) samples on host for that step. That costs the greedy rows a step's worth of device-sampling speed; it does not cost correctness. #111's rejection design avoids this entirely, at the cost of rejecting requests that this design instead admits.

**Guard added that #111 does not need.** A block-output model's per-request `SamplingParams` clone is always neutralized to a fixed, non-greedy `temperature=1.0` (`_neutralize_model_owned_sampling`), so `InputBatch.all_greedy` is always `False` for one. Combined with `supports_non_greedy_sampling_on_device=False`, every step would silently fall back to host sampling, and host sampling cannot build a block-output canvas (existing test: `test_startup_requires_device_sampling_for_block_models`). Added a startup-time `ValueError` for that combination, matching this repo's existing convention of raising when a declared capability contradicts block-output status.

**Documentation.** Updated `AGENTS.md`'s `model_capabilities` table and `_TT_PLATFORM_CONFIG_ATTRS` list, and added one sentence to `README.md`'s existing host-sampling-fallback paragraph naming this case.

AI assistance was used to review #111 and to design and implement this alternative. Posting for reviewer scrutiny, not asserting it is merge-ready as-is.

## Related Artifacts

Alternative to #111 (same incident, same new `model_capabilities` key, different enforcement point; see Details). No tracked issue number; the motivating incident is documented in #111's own description.

## Validation

Host-only; no Tenstorrent hardware in this environment. Device-side validation (does host sampling actually run correctly for a real greedy-only device-sampling model under this fallback) is outstanding and should happen before merge, same caveat #111 itself carries.

```text
PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/test_block_request_validation.py tests/test_model_runner.py -v
91 passed, 14 warnings

PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/ --ignore=tests/tt
376 passed, 14 warnings

pre-commit run --files AGENTS.md README.md src/vllm_tt_plugin/platform.py src/vllm_tt_plugin/model_runner.py tests/conftest.py tests/test_block_request_validation.py tests/test_model_runner.py
trim trailing whitespace.....Passed
fix end of files.............Passed
check for merge conflicts....Passed
ruff check....................Passed
ruff format...................Passed
```

## Checklist

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -247,6 +247,7 @@ Keys currently consumed by `src/vllm_tt_plugin/platform.py`:
 | `supports_prefix_caching` | `False` | Whether the vLLM prefix cache may be used |
 | `output_tokens_per_step` | `1` | Committed output width per step. `1` is token-at-a-time; `>1` is block-output width |
 | `supports_sample_on_device` | `False` | Opt-in for on-device sampling. A requested `sample_on_device_mode` is rejected when `False` |
+| `supports_non_greedy_sampling_on_device` | `True` | Whether an enrolled device sampler can execute temperature/top_p/top_k, not only greedy. `check_perform_device_sampling` routes a non-greedy step to host sampling when `False` |
 | `supports_async_decode` | `False` | Whether async scheduling may stay on. When `False`, the platform warns and clears `async_scheduling` |
 
 Absent keys default via `.get`. That is the live contract. Do not add
@@ -335,7 +336,8 @@ lifetime of the value:
   `check_and_update_config` configures every later test. Current names in
   `tests/conftest.py` `_TT_PLATFORM_CONFIG_ATTRS`:
   `_standard_dp_visible_device_groups`, `_standard_dp_mesh_grids`,
-  `sample_on_device_mode`, `always_compat_sampling`, `_tt_vllm_config`.
+  `sample_on_device_mode`, `supports_non_greedy_sampling_on_device`,
+  `always_compat_sampling`, `_tt_vllm_config`.
   `always_compat_sampling` is assigned at runtime inside
   `_apply_check_and_update_config`, not declared as a `ClassVar` on the class
   body.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -247,7 +247,7 @@ Keys currently consumed by `src/vllm_tt_plugin/platform.py`:
 | `supports_prefix_caching` | `False` | Whether the vLLM prefix cache may be used |
 | `output_tokens_per_step` | `1` | Committed output width per step. `1` is token-at-a-time; `>1` is block-output width |
 | `supports_sample_on_device` | `False` | Opt-in for on-device sampling. A requested `sample_on_device_mode` is rejected when `False` |
-| `supports_non_greedy_sampling_on_device` | `True` | Whether an enrolled device sampler can execute temperature/top_p/top_k, not only greedy. `check_perform_device_sampling` routes a non-greedy step to host sampling when `False` |
+| `supports_random_sampling_on_device` | `True` | Whether an enrolled device sampler can execute random sampling (temperature/top_p/top_k), not only greedy. `check_perform_device_sampling` routes a random-sampling step to host sampling when `False` |
 | `supports_async_decode` | `False` | Whether async scheduling may stay on. When `False`, the platform warns and clears `async_scheduling` |
 
 Absent keys default via `.get`. That is the live contract. Do not add
@@ -336,7 +336,7 @@ lifetime of the value:
   `check_and_update_config` configures every later test. Current names in
   `tests/conftest.py` `_TT_PLATFORM_CONFIG_ATTRS`:
   `_standard_dp_visible_device_groups`, `_standard_dp_mesh_grids`,
-  `sample_on_device_mode`, `supports_non_greedy_sampling_on_device`,
+  `sample_on_device_mode`, `supports_random_sampling_on_device`,
   `always_compat_sampling`, `_tt_vllm_config`.
   `always_compat_sampling` is assigned at runtime inside
   `_apply_check_and_update_config`, not declared as a `ClassVar` on the class
--- a/README.md
+++ b/README.md
@@ -280,7 +280,7 @@ curl http://localhost:8000/v1/completions \
   }'
 ```
 
-Requests that cannot use TT on-device sampling automatically fall back to vLLM’s host-side sampling path. This fallback is selected per batch and requires no user configuration. One case: a model whose on-device sampler only implements greedy decoding (`model_capabilities['supports_non_greedy_sampling_on_device'] = False`) falls back to host sampling for the whole batch on any step that contains a non-greedy request.
+Requests that cannot use TT on-device sampling automatically fall back to vLLM’s host-side sampling path. This fallback is selected per batch and requires no user configuration. One case: a model whose on-device sampler only implements greedy decoding (`model_capabilities['supports_random_sampling_on_device'] = False`) falls back to host sampling for the whole batch on any step that contains a random-sampling request (temperature > 0).
 
 For vision models, start the server with the correct `--model`, then send a chat
 completion request with image content. Qwen 2.5-VL models can use either a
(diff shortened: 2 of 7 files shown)


sanjaradylov (approved):
I think we can have mixed batches in `test_check_perform_device_sampling_routes...` instead of single-request batches. Also, is it guaranteed that a new batch always becomes eligible for device sampling, after mixed-batch host sampling, if it's all-greedy?

By the way, is the term "non-greedy" established in the plugin / Metal? I've seen *random* sampling also.

viktorpusTT:
Thanks for the review, all three addressed:

**Terminology.** You're right, `non_greedy` wasn't the established term. This plugin already pairs `all_greedy`/`all_random` (`input_batch.py:287-289`, `:1035-1036`) and `random_reqs` (`input_batch.py:225,401,403,475`), and `model_runner.py:2214-2216` computes the same pair. Renamed `supports_non_greedy_sampling_on_device` to `supports_random_sampling_on_device` everywhere (capability key, `TTPlatform` attribute, docs, tests) in c70ebe3.

**Mixed batches in the test.** Fair catch, the original test only ever built single-request batches, so it never actually exercised "one random row sends the *whole batch* to host," it just trusted that inference. Added `test_check_perform_device_sampling_routes_mixed_batch_to_host_when_restricted` in the same commit: a 2-request batch (one greedy, one random) against a restricted sampler asserts host fallback, and a 2-request all-greedy batch against the same restricted sampler asserts device sampling still runs.

**Does a batch become eligible for device sampling again after a mixed-batch host-sampling step, once it's all-greedy?** Yes, guaranteed, for two independent reasons:

1. `check_perform_device_sampling` (`model_runner.py:1926`) is pure and stateless: every call recomputes `input_batch.all_greedy` live from the batch's current `random_reqs` set. Nothing caches its previous return value, so as soon as the random request finishes or leaves the batch, the very next call sees `all_greedy=True` again.
2. More importantly, this exact device<->host transition is already a solved problem here, because the pre-existing host-only conditions (`bad_words`, `min_p`, structured outputs, etc.) can flip it step-to-step too. `TTAsyncDecodeCoordinator` tracks `self._previous_device_sampling` (`async_decode.py:206`) and computes `sampling_mode_changed = self._previous_device_sampling != device_sampling` (`async_decode.py:243-245`) inside `plan_decode_reload`. On a transition, `reload_inputs` and `sampling_reset` both go `True` (`:252-258`), so the device sampler's state is correctly re-established rather than resumed stale. `steady_decode_scheduler_invariants_met` (`:351`) separately refuses the fast/no-reload decode path whenever `self._previous_device_sampling is not True` (`:370`), so a transition step never silently reuses a resident decode chain that was configured for the other mode. For a model that doesn't declare `supports_async_decode`, `reload_inputs` is unconditionally `True` every step anyway (`not supports_resident_decode`, `:255`), so there's no residency state to get out of sync in the first place.

I didn't add a new test for that transition specifically since it's pre-existing behavior this PR doesn't touch, not something new I introduced. Happy to add one if you'd rather have it pinned down explicitly.

nkapreTT:
Your two objections to #111 are both correct and I'd rather fix them than defend the enforcement point: `validate_request` doesn't run under the Rust frontend for a non-block-output model, and reading a mutable `TTPlatform` class attribute per request does break the "AR engines may share a process freely" invariant. Enforcing in `check_perform_device_sampling` is the better location.

Two things block #114 for generated Quetzal, though:

**1. The fallback isn't reachable for us.** `model_runner.py:1994` and `async_decode.py:800` omit `sampling_params` when `perform_device_sampling` is False, so the adapter receives `None` and raises. Past that, `_get_output_tokens` does `logits = tt_out[rows, -1, :]` while `prefill_forward` returns a `(1,)` int32 token — `compute_logits` is `NotImplementedError` for us. So this substitutes one fatal exception for another rather than falling back. (The same is already true of the eight existing host-only conditions — not something #114 introduced.)

**2. The boolean can't express the actual constraint.** Quetzal supports temperature, full-vocabulary nucleus, top-k ≤ 64 and seeds **on device** (`serving/device_sampling.py`), and raises on `min_p`, the three penalties, logprobs, and top-k > 64. So `temperature=0, presence_penalty=0.5` is a *greedy* request that passes both PRs' checks and still kills EngineCore — greedy normalization resets `top_p`/`top_k`/`min_p` but never penalties.

Suggest: keep your enforcement point, but declare a sampling **envelope** rather than a bool, and log the per-step device-vs-host decision so a Model-CI score is attributable.

viktorpusTT:
Can you please elaborate more on your suggestion? If we have an envelope, and see a request that cannot sample on device (because the model doesn't support that), but also cannot sample on host (because the model doesn't return logits), what do you suggest we should do? I can imagine several options:
* Detect early and reject such request (what #111 originally suggested).
* Neutralize the offending sampling params, log that, and sample on device.
* Modify the model to actually return logits (I have no access to the repo, so I have zero insight whether that is possible) and sample on host. This would actually make the model honor the plugin contract.

Issue #110, opened 2026-09-03
Speculative decoding: design space and a runner-model contract (umbrella)


viktorpusTT:
## Purpose

This is the runner-model contract for speculative decoding on the TT backend.
It is normative: a tt-metal model that wants speculative decoding through
`vllm-tt-plugin` implements this surface, and the plugin implements the runner
side against it.

The contract is deliberately larger than any one model supports. Its job is to
be sufficient for models nobody has written yet: different drafters, different
verify layouts, different state kinds, different batch capabilities. A model
declares what it can do and the plugin refuses the rest at config time, loudly.

Today the plugin rejects every speculative request outright
(`assert not vllm_config.speculative_config`,
`src/vllm_tt_plugin/platform.py:1474-1476`), so nothing here is reachable yet.

Two implementations exist in tt-metal and both predate this contract: MTP for
DeepSeek-V3 (`models/demos/deepseek_v3/tt/generator.py`) and MTP for Qwen3.6
(tenstorrent/tt-metal#55548). Neither is reachable from the plugin. They are
the evidence base for section 5 and the calibration for section 8, not the
definition of the contract.

### Which path to build to

There is a second, older way to emit several tokens per step on this plugin,
and a model author has to know which one to target.

The **block-output rail** is selected by `output_tokens_per_step > 1`
(`src/vllm_tt_plugin/config.py:102`). The model owns the whole loop and
`decode_forward` returns a padded block of token ids. vLLM never learns that
speculation happened: there is no `speculative_config`, no drafts, no
acceptance, and no per-request draft policy. `TTScheduler` accounts for the
block with output placeholders, and the adaptive variant emits a block only on
a solo decode step and falls back to plain one-token decode at batch two or
more.

That rail works today and it produces real single-stream numbers. It cannot be
the production path, for a reason that no amount of plugin work fixes while the
model owns the sampler: the plugin has to discard the sampling controls it
cannot honour, and it does so with a log line rather than an error
(`_neutralize_model_owned_sampling`, warned at
`src/vllm_tt_plugin/platform.py:797-805`). A caller who sends `top_p` over an
OpenAI-compatible endpoint gets something else. Structured output is worse:
with no drafts visible to vLLM, the grammar machinery cannot mask any of the
emitted tokens.

So the split is:

- **This contract is the production path.** It is what serves the API surface:
  sampling parameters honoured exactly, because `"logits"` mode hands the rows
[...]

PR #118, opened 2026-09-10
TT spec serving: variable block output, adaptive block-output DP, discovery-join timeout

arginugaTT:
Plugin-side support for TT-native speculative models that emit a variable accepted-prefix+bonus block per step:

1. Spec model-class selection: TT_GEMMA4_SPEC routes gemma-4 to its MTP/dFlash serving classes — same checkpoint and architecture, different decode class; the default path is untouched.
2. tt_adaptive_block_output: the scheduler reserves the multi-token block placeholder only on solo decode steps; batched steps schedule 1 token/request. One server serves spec at conc-1 and plain batched baseline at conc>1, and the DP gate admits these models (each engine owns its own scheduler and solo gate — validated DP=4 on the 32-chip BH Galaxy).
3. Block-output contract unchanged and strict: exactly output_tokens_per_step valid tokens per step, model-owned sampling (sample_on_device_mode="all"), synchronous scheduling.

Validated: LB (P150x8) adaptive serving end-to-end, and BH Galaxy DP=4 (4×(1,8)) boot + benchmark sweep.

---

## Open work items (from review, vllm-tt-plugin#118 and tt-metal#56048)

**1. Allocate KV pages before executing an adaptive block.** An adaptive dFlash
step runs several internal verification iterations against vLLM-owned KV
tensors, but `Scheduler.schedule` allocates with `num_lookahead_tokens=0` and
`TTScheduler._update_after_schedule` only raises
`Request.num_output_placeholders` afterwards — which accounts for pending
output tokens without allocating pages for the current execution.
`refresh_page_tables` then pads the missing columns with zero and hands those
tables to `ttnn_packed_verify_forward`. Reserve the complete adaptive
footprint before `allocate_slots`, including verification positions and
accepted-token carry beyond the emitted width. Needs a regression that checks
allocated page coverage before the first adaptive decode and before later page
crossings; placeholder accounting alone is not sufficient.

**2. Preserve the owner's session when another request is released.**
`release_request` ignores the released slot and clears the global speculative
session, so aborting request A tears down a session owned by live request B.
B then falls to one-token baseline decode while its scheduling decision
reserved a full block, and `_update_request_with_output` rejects the width.
Make `release_request` conditional on actual session ownership and state the
same lifecycle contract on both sides — this one is paired: the ownership
check is a tt-metal change, the contract is the plugin's. Needs a regression
that prefills A and B separately, aborts A, applies runner lifecycle
processing, and asserts B's scheduled and emitted widths agree.

Not in this PR: the speculative contract stack was removed from this branch,
so it stays pure adaptive block-output. The contract rail lands separately once
async batched-baseline throughput is preserved (see the migration split agreed
on #110).

### DIFF
--- a/src/vllm_tt_plugin/config.py
+++ b/src/vllm_tt_plugin/config.py
@@ -102,6 +102,59 @@ def is_tt_block_output_model(vllm_config: "VllmConfig") -> bool:
     return get_tt_output_tokens_per_step(vllm_config) > 1
 
 
+_ADAPTIVE_BLOCK_OUTPUT_KEY = "tt_adaptive_block_output"
+
+
+def is_tt_adaptive_block_output_model(vllm_config: "VllmConfig") -> bool:
+    """Whether the model commits a block ONLY when it decodes alone (batch==1).
+
+    A plain block-output model owns a single request state and requires
+    ``max_num_seqs 1`` / no data-parallelism. An ADAPTIVE block-output model
+    emits its multi-token block only on steps that schedule exactly one decode
+    request, and falls back to plain 1-token baseline decode whenever two or
+    more requests batch together. That lifts the ``max_num_seqs 1`` and (later)
+    data-parallel restrictions: at low concurrency each request gets the block
+    speedup, at higher concurrency the server is a plain batched baseline
+    (never worse). The scheduler reserves the K-token placeholder block only for
+    a solo decode step (see TTScheduler), matching the model's batch gate.
+    """
+    additional = getattr(vllm_config, "additional_config", None) or {}
+    return bool(additional.get(_ADAPTIVE_BLOCK_OUTPUT_KEY, False))
+
+
+def store_tt_adaptive_block_output(vllm_config: "VllmConfig", flag: bool) -> None:
+    additional = getattr(vllm_config, "additional_config", None)
+    if not isinstance(additional, dict):
+        additional = {}
+        vllm_config.additional_config = additional
+    additional[_ADAPTIVE_BLOCK_OUTPUT_KEY] = bool(flag)
+
+
+_ADAPTIVE_BLOCK_MAX_PROMPT_KEY = "tt_adaptive_block_max_prompt_tokens"
+
+
+def get_tt_adaptive_block_max_prompt_tokens(vllm_config: "VllmConfig") -> int:
+    """Prompt-length frontier for the adaptive block path (0 = no limit).
+
+    An adaptive block model whose fused capture cannot fit beyond some prompt
+    length serves longer prompts as plain baseline (width-1 steps) for their
+    whole lifetime. The scheduler must reserve width-1 for those requests even
+    on solo decode steps, so the model declares the SAME frontier it gates on.
+    """
+    additional = getattr(vllm_config, "additional_config", None) or {}
+    return int(additional.get(_ADAPTIVE_BLOCK_MAX_PROMPT_KEY, 0))
+
+
+def store_tt_adaptive_block_max_prompt_tokens(
+    vllm_config: "VllmConfig", limit: int
+) -> None:
+    additional = getattr(vllm_config, "additional_config", None)
+    if not isinstance(additional, dict):
+        additional = {}
+        vllm_config.additional_config = additional
+    additional[_ADAPTIVE_BLOCK_MAX_PROMPT_KEY] = int(limit)
+
+
 def store_tt_output_tokens_per_step(
     vllm_config: "VllmConfig", output_tokens_per_step: int
 ) -> None:
--- a/src/vllm_tt_plugin/scheduler.py
+++ b/src/vllm_tt_plugin/scheduler.py
@@ -145,6 +145,10 @@ class TTScheduler(AsyncScheduler):
         # its SchedulerOutput in update_from_output so _update_request_with_output
         # sees the decision that produced THIS output (not a later schedule).
         self._committing_block_step_decisions: dict[str, bool] = {}
+        # Request id that owns the adaptive model's SINGLE speculative
+        # session, mirrored from scheduling-side facts (see
+        # _mirror_spec_session). Only its owner can emit a block.
+        self._spec_session_owner: str | None = None
         self._output_tokens_per_step = get_tt_output_tokens_per_step(self.vllm_config)
         self._is_block_output_model = is_tt_block_output_model(self.vllm_config)
         # Adaptive: emit the block only on a solo decode step; batch >1 decodes
@@ -638,6 +642,8 @@ class TTScheduler(AsyncScheduler):
         # matching SchedulerOutput back to update_from_output, so the map always
         # pairs with the output being committed.
         solo = len(scheduler_output.num_scheduled_tokens) == 1
+        if self._is_adaptive_block:
+            self._mirror_spec_session(scheduler_output, solo)
         decisions: dict[str, bool] = {}
         for req_id in scheduler_output.num_scheduled_tokens:
             request = self.requests[req_id]
@@ -656,17 +662,73 @@ class TTScheduler(AsyncScheduler):
                 is_decode = (
                     request.num_computed_tokens - scheduled >= request.num_prompt_tokens
                 )
-                spec_eligible = (
-                    self._adaptive_block_max_prompt == 0
-                    or request.num_prompt_tokens <= self._adaptive_block_max_prompt
+                # Owning the model's single spec session is part of the
+                # gate: the model serves a solo decode as plain baseline
+                # whenever it has no session for THAT request, so reserving
+                # a block for a non-owner would reserve a width the model
+                # cannot emit (see _mirror_spec_session).
+                block_step = (
+                    solo
+                    and is_decode
+                    and self._spec_eligible(request)
+                    and self._spec_session_owner == req_id
                 )
-                block_step = solo and is_decode and spec_eligible
             else:
                 block_step = True
             if block_step:
                 request.num_output_placeholders += extra_placeholders
             decisions[req_id] = block_step
         set_tt_block_step_decisions(scheduler_output, decisions)
+
+    def _spec_eligible(self, request: Request) -> bool:
+        """Prompt short enough for the model to arm a spec session for it."""
+        return (
+            self._adaptive_block_max_prompt == 0
+            or request.num_prompt_tokens <= self._adaptive_block_max_prompt
+        )
+
+    def _mirror_spec_session(self, scheduler_output, solo: bool) -> None:
+        """Track which request owns the adaptive model's SINGLE spec session.
+
+        The model keeps one global session: drafter taps are captured during a
+        request's own prefill, and only the request that owns them can emit a
+        multi-token block. Every transition of that session is driven by the
+        shape of a step, so the scheduler can mirror it exactly without asking
+        the model -- and that is what keeps the width this scheduler RESERVES
+        equal to the width the model EMITS:
+
+        * a PREFILL step re-seats the session. The model captures taps only for
+          a solo, spec-eligible prefill and drops the session on any other
+          prefill (batched, or a prompt over the capture frontier).
+        * a BATCHED decode step destroys it. The model releases the decoder and
+          serves plain baseline from then on, and it never re-arms, because
+          taps only ever come from a prefill. This is the case that made a
+          benchmark sweep fail: as concurrency drains back to one request, that
+          request is solo and eligible but no longer owns a session.
+        * a SOLO decode leaves ownership alone -- the owner bootstraps its
+          pending taps and keeps the session for the rest of its life.
+
+        A TT step is never mixed prefill+decode (docs/SCHEDULING.md), so the
+        first scheduled request classifies the whole step.
+        """
+        owner = self._spec_session_owner
+        # A finished or aborted owner no longer holds the session: the model
+        # clears it in release_request, and the id is gone from self.requests.
+        if owner is not None and owner not in self.requests:
+            owner = None
+        scheduled_tokens = scheduler_output.num_scheduled_tokens
+        first_id = next(iter(scheduled_tokens), None)
+        if first_id is not None:
+            request = self.requests[first_id]
+            step_is_decode = (
+                request.num_computed_tokens - scheduled_tokens[first_id]
+                >= request.num_prompt_tokens
+            )
+            if not step_is_decode:
+                owner = first_id if (solo and self._spec_eligible(request)) else None
+            elif not solo:
+                owner = None
+        self._spec_session_owner = owner
 
     def update_from_output(self, scheduler_output, model_runner_output):
         """Bind this step's block-step decisions before the base loop commits
--- a/src/vllm_tt_plugin/scheduler.py
+++ b/src/vllm_tt_plugin/scheduler.py
@@ -43,6 +43,31 @@ def get_tt_forced_reset_discard_counts(
     return dict(getattr(scheduler_output, _TT_FORCED_RESET_DISCARD_COUNTS_ATTR, {}))
 
 
+# Per-step block-output reconciliation decisions, keyed by request id. Attached
+# to the ``SchedulerOutput`` -- the ONLY object that re-associates a step's
+# output with the scheduling decision that produced it: under async scheduling
+# ``schedule()`` for step K+1 (which mutates per-``Request`` state) runs BEFORE
+# ``update_from_output`` commits step K, so a single overwritten ``Request``
+# slot would hand step K's commit the K+1 decision. The engine core holds each
+# in-flight step's ``SchedulerOutput`` in its batch queue and passes it back to
+# ``update_from_output``, so a map carried here always matches the output being
+# committed (in sync mode the pairing is trivially the same step). Mirrors the
+# forced-reset-count pattern above.
+_TT_BLOCK_STEP_DECISIONS_ATTR = "_tt_block_step_decisions"
+
+
+def set_tt_block_step_decisions(
+    scheduler_output: SchedulerOutput, decisions: dict[str, bool]
+) -> None:
+    setattr(scheduler_output, _TT_BLOCK_STEP_DECISIONS_ATTR, dict(decisions))
+
+
+def get_tt_block_step_decisions(
+    scheduler_output: SchedulerOutput,
+) -> dict[str, bool]:
+    return dict(getattr(scheduler_output, _TT_BLOCK_STEP_DECISIONS_ATTR, {}))
+
+
 class TTSchedulingMode(Enum):
     DEFAULT = "default"
     DECODE_ONLY = "decode_only"
@@ -116,6 +141,10 @@ class TTScheduler(AsyncScheduler):
         super().__init__(*args, **kwargs)
         self._forced_mode = TTSchedulingMode.DEFAULT
         self._pending_forced_reset_discard_counts: dict[str, int] = {}
+        # Block-step decisions for the step currently being committed, read off
+        # its SchedulerOutput in update_from_output so _update_request_with_output
+        # sees the decision that produced THIS output (not a later schedule).
+        self._committing_block_step_decisions: dict[str, bool] = {}
         self._output_tokens_per_step = get_tt_output_tokens_per_step(self.vllm_config)
         self._is_block_output_model = is_tt_block_output_model(self.vllm_config)
         # Adaptive: emit the block only on a solo decode step; batch >1 decodes
@@ -603,10 +632,13 @@ class TTScheduler(AsyncScheduler):
         #     and emits the single anchor). num_computed reflects prior committed
         #     steps here, so it is 0<prompt on the prefill and >=prompt on the
         #     first decode -- robust where an output-token count can still read 0.
-        # ``_tt_block_step`` records the decision so the reconciliation in
-        # _update_request_with_output takes the same path (block serving is
-        # synchronous, so the stamp and its commit never interleave).
+        # The decision is attached to THIS step's SchedulerOutput (not a Request
+        # slot): under async the next step's schedule() overwrites Request state
+        # before this step's output commits, but the engine core hands the
+        # matching SchedulerOutput back to update_from_output, so the map always
+        # pairs with the output being committed.
         solo = len(scheduler_output.num_scheduled_tokens) == 1
+        decisions: dict[str, bool] = {}
         for req_id in scheduler_output.num_scheduled_tokens:
             request = self.requests[req_id]
             if request.is_prefill_chunk:
@@ -633,7 +665,21 @@ class TTScheduler(AsyncScheduler):
                 block_step = True
             if block_step:
                 request.num_output_placeholders += extra_placeholders
-            request._tt_block_step = block_step
+            decisions[req_id] = block_step
+        set_tt_block_step_decisions(scheduler_output, decisions)
+
+    def update_from_output(self, scheduler_output, model_runner_output):
+        """Bind this step's block-step decisions before the base loop commits
+        its outputs, so ``_update_request_with_output`` reads the decision that
+        produced THIS output rather than a later schedule's overwrite."""
+        if self._is_block_output_model:
+            self._committing_block_step_decisions = get_tt_block_step_decisions(
+                scheduler_output
+            )
+        try:
+            return super().update_from_output(scheduler_output, model_runner_output)
+        finally:
+            self._committing_block_step_decisions = {}
 
     def _update_request_with_output(
         self, request: Request, new_token_ids: list[int]
@@ -641,13 +687,13 @@ class TTScheduler(AsyncScheduler):
         """Commit one block and reconcile its full physical reservation."""
         if not self._is_block_output_model:
             return super()._update_request_with_output(request, new_token_ids)
-        # Adaptive: a request that decoded batched this step committed a single
-        # baseline token (no block was reserved for it) -> plain reconciliation.
-        # The flag is stamped by _ensure_block_output_placeholders in the SAME
-        # step this output belongs to (block-output serving is synchronous, so
-        # commits never interleave with the next step's scheduling decision).
+        # Adaptive: a request that decoded batched (or over-frontier) this step
+        # committed a single baseline token (no block was reserved) -> plain
+        # reconciliation. The decision is read from THIS step's SchedulerOutput
+        # (set in _update_after_schedule), which the engine core pairs with this
+        # output even when async scheduling has already run a later schedule().
         if self._is_adaptive_block:
-            block_step = getattr(request, "_tt_block_step", None)
+            block_step = self._committing_block_step_decisions.get(request.request_id)
             if block_step is None:
                 raise RuntimeError(
                     "adaptive block-output request committed output without a "
@@ -656,15 +702,24 @@ class TTScheduler(AsyncScheduler):
             if not block_step:
                 return super()._update_request_with_output(request, new_token_ids)
         if request.async_tokens_to_discard:
+            # A block step reserved K placeholders; the AsyncScheduler discard
+            # path drains only one per stale frame, so it cannot balance a
+            # dropped block. Block requests are never reset-preempted
+            # (reset_prefix_cache raises while one runs) and solo spec steps are
+            # not KV-preempted, so this must not happen -- fail loudly rather
+            # than silently leak placeholders.
             raise RuntimeError(
-                "A stale async output reached synchronous block serving; "
-                "block-output async scheduling and running prefix resets are "
-                "unsupported"
+                "A stale async output reached block serving for a block step; "
+                "block-output frames cannot be discarded (reset/preempt of a "
+                "running block request is unsupported)"
             )
         if len(new_token_ids) != self._output_tokens_per_step:
             raise ValueError(
                 "Model output width violates output_tokens_per_step: "
-                f"{len(new_token_ids)} != {self._output_tokens_per_step}"
+                f"{len(new_token_ids)} != {self._output_tokens_per_step} "
+                f"(req_id={request.request_id!r}); the scheduler reserved a "
+                "block but the model returned a different width -- the "
+                "scheduler and model block gates disagree"
             )
 
         # Scheduler appends token-by-token and trims at EOS, stop tokens,
--- a/src/vllm_tt_plugin/scheduler.py
+++ b/src/vllm_tt_plugin/scheduler.py
@@ -12,7 +12,9 @@ from vllm.v1.core.sched.scheduler import Scheduler
 from vllm.v1.request import Request, RequestStatus
 
 from vllm_tt_plugin.config import (
+    get_tt_adaptive_block_max_prompt_tokens,
     get_tt_output_tokens_per_step,
+    is_tt_adaptive_block_output_model,
     is_tt_block_output_model,
 )
 from vllm_tt_plugin.logger import init_tt_logger
@@ -116,6 +118,15 @@ class TTScheduler(AsyncScheduler):
         self._pending_forced_reset_discard_counts: dict[str, int] = {}
         self._output_tokens_per_step = get_tt_output_tokens_per_step(self.vllm_config)
         self._is_block_output_model = is_tt_block_output_model(self.vllm_config)
+        # Adaptive: emit the block only on a solo decode step; batch >1 decodes
+        # as plain baseline. Lets max_num_seqs>1 coexist with block-output.
+        self._is_adaptive_block = is_tt_adaptive_block_output_model(self.vllm_config)
+        # Prompt-length frontier for the block path (0 = none): a longer prompt
+        # is served as plain baseline by the model for its whole lifetime, so
+        # its steps reserve width-1 even when solo.
+        self._adaptive_block_max_prompt = get_tt_adaptive_block_max_prompt_tokens(
+            self.vllm_config
+        )
         if self._is_block_output_model:
             assert self.num_sampled_tokens_per_step == 1, (
                 "Block-output accounting requires upstream to reserve exactly "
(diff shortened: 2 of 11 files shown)


viktorpusTT:
## 1. A solo decode step scheduled for a request that does not own the speculative session makes `Gemma4DFlashForCausalLM` emit a full block against a single reserved placeholder

The failure sequence:

1. Request A and request B are both in `TTScheduler.running`. Request A owns the
   session, so `TTScheduler._mirror_spec_session` holds
   `self._spec_session_owner == "A"`. On the tt-metal side,
   `Gemma4DFlashForCausalLM._spec_bootstrap` has already run for request A, so
   `Gemma4DFlashForCausalLM._spec_active` is `True`,
   `Gemma4DFlashForCausalLM._spec_pending` is `None`, and
   `Gemma4DFlashForCausalLM._spec_pending_owner` is `None`
   (`models/demos/gemma4/tt/generator_vllm.py`:2242 clears both).
2. Request A reaches `max_tokens`. Upstream `Scheduler.schedule` skips it at
   `.venv/lib/python3.12/site-packages/vllm/v1/core/sched/scheduler.py`:473 to 482.
   That skip is guarded on `request.num_output_placeholders > 0`, which is an
   async-scheduling condition: the comment in upstream vLLM reads "Async
   scheduling: Avoid scheduling an extra step when we are sure that the previous
   step has reached request.max_tokens." This change is what admits async
   scheduling for an adaptive model, at `src/vllm_tt_plugin/platform.py`:1840.
3. `TTScheduler.schedule` therefore produces a step whose
   `scheduler_output.num_scheduled_tokens` contains only request B.
4. `TTScheduler._update_after_schedule` computes `solo = True` at
   `src/vllm_tt_plugin/scheduler.py`:644. It calls
   `TTScheduler._mirror_spec_session`, which sees a solo decode step and takes the
   "leave the owner alone" path, so `self._spec_session_owner` stays `"A"`.
5. `TTScheduler._update_after_schedule` evaluates `block_step` for request B at
   `src/vllm_tt_plugin/scheduler.py`:670. `self._spec_session_owner == req_id` is
   false, so `block_step` is `False` and exactly one placeholder is reserved for
   request B.
6. `TTModelRunner` executes the step. `Gemma4DFlashForCausalLM.decode_forward`
   sees `batch == 1`. `Gemma4DFlashForCausalLM._spec_pending` is `None`, so the
   ownership check at `models/demos/gemma4/tt/generator_vllm.py`:2441
   (`if self._spec_pending is not None and not self._spec_pending_is_mine(...)`)
   is skipped entirely. `Gemma4DFlashForCausalLM._spec_active` is `True`, so the
   guard at `models/demos/gemma4/tt/generator_vllm.py`:2462
   (`if not self._spec_active:`) does not fire either.
7. `Gemma4DFlashForCausalLM.decode_forward` calls
   `dec.refresh_page_tables(row)` at
   `models/demos/gemma4/tt/generator_vllm.py`:2487, re-pointing request A's fused
   decoder at request B's page table, and runs the block loop. It emits
   `output_tokens_per_step` tokens for request B, speculated from request A's
   residual taps. An anchor mismatch produces only a warning at
   `models/demos/gemma4/tt/generator_vllm.py`:2475, whose text is "trusting the
   session (advisory-input contract)".
8. `TTScheduler._update_request_with_output` reads `block_step` as `False` for
   request B and returns `super()._update_request_with_output(request, new_token_ids)`
   at `src/vllm_tt_plugin/scheduler.py`:765, with no width check.
   `AsyncScheduler._update_request_with_output` appends all
   `output_tokens_per_step` tokens, then executes
   `request.num_output_placeholders -= len(new_token_ids)` against a reservation of
   1, then hits its bare `assert request.num_output_placeholders >= 0`.

The result is an opaque `AssertionError` raised from inside upstream vLLM with no
request id and no widths, and, under `python -O`, no error at all plus a negative
`request.num_output_placeholders`. The committed tokens are wrong in either case,
because they were speculated from request A's prompt residuals.

Suggestion. Add the fourth transition to `TTScheduler._mirror_spec_session`: a
solo decode step whose single scheduled request is not
`self._spec_session_owner` must clear `self._spec_session_owner`. That alone does
not fix the emitted width, because `Gemma4DFlashForCausalLM.decode_forward` has no
ownership check on an active session. The tt-metal side needs an equivalent
identity check on the active-session path, comparing
`Gemma4DFlashForCausalLM._spec_pt_identity(page_table)` against the identity
recorded when the session was bootstrapped, and falling back to plain baseline on
a mismatch.

Consider replacing _mirror_spec_session with a model-side query, the way `release_request` and `release_persistent_capture` already are per `AGENTS.md:268`. That deletes the whole class of desync bugs.

## 2. `TTScheduler._spec_eligible` measures the prompt-length frontier in a different quantity than `Gemma4DFlashForCausalLM.prefill_forward` does, so a preempted and resumed request crosses the frontier on the model side only

Both sides must measure the same quantity or the reserved width and the emitted
width diverge. They do not.

1. `TTScheduler._spec_eligible` at `src/vllm_tt_plugin/scheduler.py`:683 to 688
   compares `request.num_prompt_tokens` against
   `self._adaptive_block_max_prompt`. `Request.num_prompt_tokens` is the length of
   the original prompt and never changes as the request generates tokens.
2. `Gemma4DFlashForCausalLM.prefill_forward` at
   `models/demos/gemma4/tt/generator_vllm.py`:2212 to 2214 computes
   `_n0 = int(_pl[0])` from the `prompt_lens` keyword argument, and compares `_n0`
   against `GEMMA4_DFLASH_MAX_SPEC_ISL`.
3. `TTModelRunner._prepare_model_inputs` computes that `prompt_lens` value at
   `src/vllm_tt_plugin/model_runner.py`:1212 as
   `prompt_lens = input_positions + chunk_lens`. Its own comment at
   `src/vllm_tt_plugin/model_runner.py`:1204 to 1206 states the consequence:
   "Resumed from preemption: the chunk spans the generated output tokens too, so
   the full sequence is replayed to rebuild KV."

So on a resumed prefill, `Gemma4DFlashForCausalLM.prefill_forward` measures
prompt tokens plus generated tokens, while `TTScheduler._spec_eligible` measures
prompt tokens alone.

The failure sequence:

1. Request A arrives with a prompt of N tokens, where N is at or below
   `GEMMA4_DFLASH_MAX_SPEC_ISL`. `TTScheduler._mirror_spec_session` seats request A
   as `self._spec_session_owner`, and request A serves blocks.
2. KV pressure builds. Upstream `Scheduler._preempt_request` at
   `.venv/lib/python3.12/site-packages/vllm/v1/core/sched/scheduler.py`:1212 to
   1235 preempts request A, sets `request.num_computed_tokens = 0`, and pushes it
   back to `TTScheduler.waiting`. Preemption is reachable only because this change
   lifts the `max_num_seqs == 1` rule at `src/vllm_tt_plugin/platform.py`:1784.
   By then request A has generated G tokens, and N plus G exceeds
   `GEMMA4_DFLASH_MAX_SPEC_ISL`.
3. `TTScheduler.schedule` resumes request A as a solo prefill step that replays
   all N plus G tokens.
4. `TTScheduler._mirror_spec_session` classifies the step as a prefill, sees
   `solo` true, and calls `TTScheduler._spec_eligible(request)`, which compares
   `request.num_prompt_tokens`, still N, against the frontier and returns `True`.
   `TTScheduler._mirror_spec_session` therefore seats request A as
   `self._spec_session_owner`.
5. `Gemma4DFlashForCausalLM.prefill_forward` receives `prompt_lens` of N plus G,
   finds `_n0 > _max_spec_isl`, calls
   `Gemma4DFlashForCausalLM._spec_drop_session("prompt over spec ceiling")`, and
   returns the plain baseline prefill. No drafter taps are captured.
6. On the next step `TTScheduler._update_after_schedule` evaluates `block_step`
   for request A: `solo` is true, `is_decode` is true,
   `TTScheduler._spec_eligible` is true, and `self._spec_session_owner == req_id`
   is true. It reserves `output_tokens_per_step` placeholders.
7. `Gemma4DFlashForCausalLM.decode_forward` finds
   `Gemma4DFlashForCausalLM._spec_pending` is `None` and
   `Gemma4DFlashForCausalLM._spec_active` is `False`, takes the branch at
   `models/demos/gemma4/tt/generator_vllm.py`:2462, and returns one token.
8. `TTScheduler._update_request_with_output` finds `block_step` is `True` and
   `len(new_token_ids)` is 1, and raises the `ValueError` at
   `src/vllm_tt_plugin/scheduler.py`:778. The engine core dies.

Suggestion. Make `TTScheduler._spec_eligible` compare the same quantity the model
compares. On the scheduling side that quantity is `request.num_tokens`, which
includes generated tokens, not `request.num_prompt_tokens`. Alternatively, record
on the request whether its most recent prefill step armed a session, and gate on
that recorded fact instead of recomputing the frontier.

## 3. `TT_MODEL_CLASS_OVERRIDES` never takes effect for the architecture names it documents

Two independent defects stop it working.

1. `_register_model_if_missing` registers only when the architecture is absent
   from the registry. Its body at `src/vllm_tt_plugin/platform.py`:556 to 557 is
   `if model_arch not in ModelRegistry.get_supported_archs(): ModelRegistry.register_model(model_arch, model_path)`.
   `ModelRegistry.get_supported_archs()` returns every architecture upstream vLLM
   ships, not only those the plugin added. In the installed vLLM 0.26.0 in this
   repository's `.venv`, the global `ModelRegistry` is built from `_VLLM_MODELS`
   and already contains 362 architectures at process start, including
   `Gemma4ForCausalLM`. An override naming any of them is discarded with no log,
   no warning and no error.
2. `TTPlatform.check_and_update_config` rewrites the checkpoint's architecture
   names in place at `src/vllm_tt_plugin/platform.py`:1578 to 1580, prefixing
   every name that does not already start with `TT`. So the architecture vLLM
   actually resolves for a Gemma4 checkpoint is `TTGemma4ForCausalLM`, and an
   override keyed on `Gemma4ForCausalLM` is never consulted even if it had been
   registered.

The docstring example at `src/vllm_tt_plugin/platform.py`:566 uses the unprefixed
name `Gemma4ForCausalLM`, so it is defeated by both defects. The operator sets the
variable, sees no diagnostic, and serves the default class.

Suggestion. Register override entries with `ModelRegistry.register_model`
unconditionally rather than through `_register_model_if_missing`, since upstream
vLLM permits re-registration and logs it. Register each entry under both the
given name and its `TT`-prefixed form, or require the `TT`-prefixed form and say
so in the docstring. Log every applied override at INFO, the way
`_register_models_from_extra_dir` already does at
`src/vllm_tt_plugin/platform.py`:1032 to 1037. Raise when an override names an
architecture that cannot be made authoritative.

Or better: Split TT_MODEL_CLASS_OVERRIDES out. It is unrelated to adaptive block output, undocumented, and non-functional. It should not ride along.

## 4. `TT_MODEL_CLASS_OVERRIDES` is documented in no operator-facing file, although it outranks both documented registration mechanisms

`README.md` has a dedicated section, "Registering models dynamically
(`EXTRA_MODELS_DIR`)", at `README.md`:439, which describes the bundle mechanism
and the `TT_VLLM_BUILTIN_MODELS=0` switch at `README.md`:458. That is where the
precedence order belongs. Searching the head commit for the string
`TT_MODEL_CLASS_OVERRIDES` returns only `src/vllm_tt_plugin/platform.py` and
`tests/test_model_class_overrides.py`. There is no mention in `README.md`, in
`AGENTS.md`, or in any file under `docs/`.

A second caveat also needs stating. `Gemma4ForCausalLM` is one of six architecture
strings registered to the same target at
`src/vllm_tt_plugin/platform.py`:1219 to 1227, so an operator must override the
specific architecture string the checkpoint configuration resolves to, and
overriding one alias leaves the other five on the default class.

Suggestion. Add `TT_MODEL_CLASS_OVERRIDES` to `README.md` immediately after the
`EXTRA_MODELS_DIR` section: the value format, the exact architecture-name form
that takes effect after the `TT`-prefix rewrite at
`src/vllm_tt_plugin/platform.py`:1578 to 1580, the precedence order, and the alias
caveat. Document it alongside the other registration environment variables an
operator sets, which are `EXTRA_MODELS_DIR`, `TT_VLLM_BUILTIN_MODELS`,
`TT_LLAMA_TEXT_VER`, `TT_QWEN3_TEXT_VER` and `TT_QWEN35_TEXT_VER`.

## 5. `docs/SCHEDULING.md` now makes two false statements about block-output models

1. `docs/SCHEDULING.md`:111 to 112 reads "Block-output models are stricter: the
   platform refuses to start unless `--no-async-scheduling` is passed." That is
   now false for a model declaring `tt_adaptive_block_output`. The gate at
   `src/vllm_tt_plugin/platform.py`:1840 to 1843 is
   `if vllm_config.scheduler_config.async_scheduling and not adaptive_block_output`,
   so an adaptive block-output model starts with async scheduling on.
2. The section "Block-output reservation" at `docs/SCHEDULING.md`:193 to 205 says
   a "synchronous block-output model reserves one complete output block through
   the same accounting" and that "When `output_tokens_per_step > 1`, the base
   scheduler reserves its normal sampled-token placeholder and `TTScheduler`
   reserves the remaining physical block width." For an adaptive model,
   `TTScheduler._update_after_schedule` adds the extra placeholders only when
   `solo and is_decode and self._spec_eligible(request) and self._spec_session_owner == req_id`,
   at `src/vllm_tt_plugin/scheduler.py`:670 to 679. Every other step reserves one
   placeholder. The section also does not mention the per-step decision map
   attached to each `SchedulerOutput` at `src/vllm_tt_plugin/scheduler.py`:46 to
   68, which is the mechanism that makes async block serving safe.

Suggestion. Update `docs/SCHEDULING.md`:111 to 112 to name
`tt_adaptive_block_output` as the exception. Extend the "Block-output reservation"
section with the four conditions the adaptive reservation depends on, and with the
`_tt_block_step_decisions` attribute carried on `SchedulerOutput`, placed next to
the existing description of the forced-reset discard counts.

## 6. The `AGENTS.md` capability table understates what `tt_adaptive_block_output` relaxes and omits every validation on both new keys

The new row for `tt_adaptive_block_output` at `AGENTS.md`:251 says only "Relaxes
the `max_num_seqs 1` and DP gates". `TTPlatform.check_and_update_config` relaxes
three more gates for the same key: the distributed executor backend gate adds
`"mp"` to `allowed_backends` at `src/vllm_tt_plugin/platform.py`:1813 to 1819, the
async scheduling gate stops raising at
`src/vllm_tt_plugin/platform.py`:1840 to 1843, and the block-output sampling gate
accepts `"decode_only"` in addition to `"all"` at
`src/vllm_tt_plugin/platform.py`:1906 to 1917. The row also omits that
`src/vllm_tt_plugin/platform.py`:1653 to 1656 raises
`ValueError("tt_adaptive_block_output requires output_tokens_per_step > 1")`.

The row for `tt_adaptive_block_max_prompt_tokens` at `AGENTS.md`:252 omits both of
its validations: `src/vllm_tt_plugin/platform.py`:1664 to 1668 raises when the
value is negative, and `src/vllm_tt_plugin/platform.py`:1669 to 1672 raises
`"tt_adaptive_block_max_prompt_tokens requires tt_adaptive_block_output"` when a
non-zero value is declared without the adaptive key.

A tt-metal model author reading only this table will declare a capability set the
platform rejects at startup.

Suggestion. Extend `AGENTS.md`:251 to list all five relaxed gates and the
`output_tokens_per_step > 1` requirement, and extend `AGENTS.md`:252 with both
raises.

## 7. The two new `additional_config` keys break the convention documented in the same file for platform-derived keys

This change adds `_ADAPTIVE_BLOCK_OUTPUT_KEY = "tt_adaptive_block_output"` at
`src/vllm_tt_plugin/config.py`:105 and
`_ADAPTIVE_BLOCK_MAX_PROMPT_KEY = "tt_adaptive_block_max_prompt_tokens"` at
`src/vllm_tt_plugin/config.py`:133, into the same top-level dictionary with no
leading underscore. They read as user input sitting beside
`_tt_resolved_lane_count`. An operator who passes
`--additional-config '{"tt_adaptive_block_output": true}'` has it silently
overwritten by `src/vllm_tt_plugin/platform.py`:1657, with no error and no log.

Two further inconsistencies in the same block. `store_tt_adaptive_block_output` at
`src/vllm_tt_plugin/config.py`:125 and
`store_tt_adaptive_block_max_prompt_tokens` at
`src/vllm_tt_plugin/config.py`:148 carry no docstring, while every other accessor
and writer in the file has one. `store_tt_adaptive_block_max_prompt_tokens`
performs no validation, while its sibling `store_tt_output_tokens_per_step` at
`src/vllm_tt_plugin/config.py`:162 to 170 rejects non-integers and values below 1.
The validation for the new key lives in `src/vllm_tt_plugin/platform.py`:1664 to
1672 instead, so two sibling knobs validate in two different files.

Suggestion. Rename the keys to `_tt_adaptive_block_output` and
`_tt_adaptive_block_max_prompt_tokens` to match
`src/vllm_tt_plugin/config.py`:40 to 41. Add docstrings to both writers stating
they are internal handoff and not user-facing. Move the range check into
`store_tt_adaptive_block_max_prompt_tokens`.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -581,27 +625,163 @@ def _update_after_schedule(self, scheduler_output: SchedulerOutput) -> None:
...
+        # other step (batched decodes, and the prefill that host-samples the
+        # anchor) commits exactly one plain token and reserves one placeholder.
+        # This MUST match the model's own gate, which runs the spec block in
+        # decode_forward when batch == 1 and a plain token otherwise:
+        #   - solo == batch 1 (exactly one request scheduled this step);
+        #   - decode == the prompt is fully computed (a prefill step still has
+        #     num_computed_tokens < num_prompt_tokens, so it takes prefill_forward
+        #     and emits the single anchor). num_computed reflects prior committed
+        #     steps here, so it is 0<prompt on the prefill and >=prompt on the
+        #     first decode -- robust where an output-token count can still read 0.
```
viktorpusTT:
```suggestion
        #   - decode == the prompt was fully computed BEFORE this step (see the
        #     decode test below, which subtracts this step's scheduled tokens).
```

src/vllm_tt_plugin/scheduler.py
viktorpusTT:
```suggestion
        """Reserve the physical output each scheduled step will commit.

        The platform removes the upstream diffusion marker, so vLLM reserves its
        normal one sampled-token placeholder. A plain block-output model commits
        a K-token canvas on every step, so reserve the remaining K-1 positions.
        An adaptive block-output model commits K only on a step that satisfies
        the ``block_step`` predicate below, and one token on every other step,
        so the extra reservation is per step. Each step's decision is recorded
        on its own SchedulerOutput because update_from_output must reconcile a
        commit against the decision that produced it.
        """
```

src/vllm_tt_plugin/config.py
```diff
@@ -102,6 +102,59 @@ def is_tt_block_output_model(vllm_config: "VllmConfig") -> bool:
...
+
+    A plain block-output model owns a single request state and requires
+    ``max_num_seqs 1`` / no data-parallelism. An ADAPTIVE block-output model
+    emits its multi-token block only on steps that schedule exactly one decode
+    request, and falls back to plain 1-token baseline decode whenever two or
+    more requests batch together. That lifts the ``max_num_seqs 1`` and (later)
+    data-parallel restrictions: at low concurrency each request gets the block
+    speedup, at higher concurrency the server is a plain batched baseline
+    (never worse). The scheduler reserves the K-token placeholder block only for
+    a solo decode step (see TTScheduler), matching the model's batch gate.
```
viktorpusTT:
```suggestion
    more requests batch together. That lifts the ``max_num_seqs 1`` and
    data-parallel restrictions: at low concurrency each request gets the block
    speedup, at higher concurrency the server is a plain batched baseline
    (never worse). ``TTScheduler._update_after_schedule`` owns the reservation
    predicate and reserves the K-token placeholder block only when the step is
    solo, is a decode, is within ``tt_adaptive_block_max_prompt_tokens``, and
    the request owns the model's speculative session.
```

arginugaTT:
All seven corrected, pushed as `0703193`. Details where I did something other than the literal suggestion.

**1. Session hand-off.** Fixed on both sides, and your failure sequence holds exactly as written. The fourth transition is in `_mirror_spec_session`: a solo decode step whose single request is not the owner clears ownership. As you noted that alone does not fix the emitted width, so the tt-metal side now records the owner at `_spec_bootstrap` and compares it per step (`_spec_active_is_mine`), releasing the session and serving plain baseline on a mismatch -- `_spec_pending_owner` only ever covered the window before bootstrap, which is why neither existing guard fired (tt-metal#56048, `01ce718d9d4`). New tests: `test_adaptive_solo_decode_by_a_non_owner_drops_the_session` and its owner-side companion here, plus `test_live_session_is_refused_for_a_request_that_does_not_own_it` in tt-metal.

On replacing `_mirror_spec_session` with a model-side query: agreed that it deletes the class rather than this instance of it, and I would rather do it that way. I have not, in this change, because the query has to answer "will you emit a block for THIS request on THIS step" before the step runs, and the model's answer currently depends on state the runner sets up after the scheduler has committed its reservation. That is a contract change, not a refactor, so I would rather land it against the contract rail than bolt it onto the block rail.

**2. Frontier quantity.** Took your alternative rather than the first suggestion, because `request.num_tokens` fixes the resumed-prefill case but introduces a worse one: it grows on every decode step, so a request whose prompt sits just under the frontier flips to width-1 reservation mid-generation while the model keeps emitting blocks -- the same engine kill, reached by an ordinary long generation instead of a preemption. The frontier is a property of the prefill that armed the session, so it is now applied once, in `_mirror_spec_session`, against `request.num_computed_tokens` -- which at that point is exactly the `prompt_lens` the model measures, for a fresh prefill, a chunk, and a resumed replay alike. `block_step` no longer re-derives it; ownership carries it. `_spec_eligible` is now `_spec_frontier_ok(measured_len)` so it cannot silently be handed a request again.

**3. `TT_MODEL_CLASS_OVERRIDES`.** Both defects confirmed and fixed: registration is unconditional (upstream permits re-registration), under both the given and the `TT`-prefixed name, logged at INFO like `_register_models_from_extra_dir`. Worth recording why it appeared to work for us: the override loop runs before the built-in map, so the `TT`-prefixed names our launches pass were genuinely absent at that moment and did register. Every example in the docstring used the bare name, which is defeated by both defects -- so the mechanism was non-functional exactly as documented, which is the part that matters for anyone else. New test registers over an architecture already in the registry, which is what the old code discarded.

**4.** Documented in `README.md` after the `EXTRA_MODELS_DIR` section: the name form that takes effect after the `TT`-prefix rewrite, the alias caveat, the precedence order, and the INFO line to grep for when it appears not to work.

**5, 6, 7.** Corrected as suggested. `docs/SCHEDULING.md` gets the async exception and a new "Adaptive block-output reservation" subsection with the four conditions and the `_tt_block_step_decisions` map; `AGENTS.md` lists all five relaxed gates and both keys' raises; the keys are `_tt_adaptive_block_output` / `_tt_adaptive_block_max_prompt_tokens`, both writers have docstrings naming them internal handoff, and the range check moved next to `store_tt_output_tokens_per_step`. The pairing rule (non-zero frontier requires the adaptive key) stays in the platform, since only the platform sees both capabilities.

**On splitting `TT_MODEL_CLASS_OVERRIDES` out**: I agree it is unrelated to adaptive block output and should not ride along. I have not split it yet because it is load-bearing for the tt-inference-server specs that select the dFlash and MTP serving classes, so a split needs both PRs in flight, not one. Happy to do it either way -- your call on whether you want it as its own PR before this one merges, or extracted after.

**On the contract rail**: 473 host tests pass here, and tt-metal now declares the four #125 admission keys on a production model (`Gemma4DFlashForCausalLM`), which I believe closes the AGENTS.md section 6 pairing gate on #125 without needing the `models/vllm_test_utils/` stub to be the only declarer. Ping me if you want me to exercise #125+#127+#128+#130 stacked against it on a P150x8 -- I have the box warm and the 31B dFlash sessions working on the block rail, so the comparison would be apples to apples.

viktorpusTT:
The latest [#131](https://github.com/tenstorrent/vllm-tt-plugin/pull/131)/[#56646](https://github.com/tenstorrent/tt-metal/pull/56646) tests exercised the model-owned proposer path through a real vLLM server: proposal, scheduler round-trip, zero/partial/full greedy acceptance, and opaque hidden-handle handoff. The dummy model does no device compute. Its reported single-request framework cost was about 1.3 ms per iteration at K=5, approximately flat whether the iteration committed one token or six.

This gives us a concrete contract path to try with Gemma4. It does not establish Gemma4's cost on Blackhole or reproduce the earlier host-overhead measurement, so I suggest a matched real-model comparison rather than treating long internal blocks as a permanent requirement.

I suggest to: split `TT_MODEL_CLASS_OVERRIDES` if registration needs it independently. Keep adaptive block-output support separate and freeze further scheduler/session features while the contract adapter proceeds. I would hold the adaptive portion unless an immediate deployment needs a bounded interim version; any interim merge still needs paired speculative correctness and lifecycle evidence. The current dFlash fill/truncate behavior needs resolution before the combined serving path is treated as correct.

The contract stack does not depend on #118. Let's use #118 as a pinned comparison path and retire model-specific block-output support when the corresponding contract adapter passes its acceptance tests.

arginugaTT:
@viktorpusTT 

I did you #118 contract work on my implementation, I need async support on dFlash for my adaptive engine to work at higher concurrencies (using baseline on conc>1 and spec decode for conc=1) for overall best server implementation.

My current block trail approach has async support.

viktorpusTT:
**1. Allocate KV pages before executing an adaptive block**

Adaptive dFlash serving performs multiple internal verification iterations during one vLLM decode step. Those iterations use vLLM-owned key/value (KV) cache tensors. `Request.num_output_placeholders` accounts for pending output tokens, but increasing `Request.num_output_placeholders` after scheduling does not allocate KV pages for the current execution.

The failure mechanism is:

1. vLLM `Scheduler.schedule` schedules the first decode after a prefill anchor with `num_new_tokens=1`. `Scheduler.schedule` calls `kv_cache_manager.allocate_slots` with `num_lookahead_tokens=0`. See [upstream scheduler.py:564-568](https://github.com/vllm-project/vllm/blob/v0.26.0/vllm/v1/core/sched/scheduler.py#L564-L568).
2. `TTScheduler._update_after_schedule` subsequently increases `Request.num_output_placeholders` to the adaptive block width. `TTScheduler._update_after_schedule` does not allocate additional KV pages.
3. `Gemma4DFlashForCausalLM.decode_forward` supplies the current request page table to `DFlashFusedDecoder.refresh_page_tables`, then repeatedly calls `DFlashFusedDecoder.step` until the output block is full. See [generator_vllm.py:3053-3059](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/generator_vllm.py#L3053-L3059) and [generator_vllm.py:3080-3098](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/generator_vllm.py#L3080-L3098).
4. `DFlashFusedDecoder.refresh_page_tables` copies the allocated block IDs and pads missing columns with zero. `DFlashFusedDecoder` passes those page tables and the runner-owned KV tensors to `ttnn_packed_verify_forward`. See [dflash_drafter.py:1847-1854](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/dflash_drafter.py#L1847-L1854) and [dflash_drafter.py:1264-1273](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/dflash_drafter.py#L1264-L1273).

Requested change: Reserve KV capacity before `allocate_slots` for the complete adaptive execution footprint, including verification positions and accepted-token carry beyond the emitted width. Add a regression that checks allocated page coverage before the first adaptive decode and before later page crossings. Output-placeholder accounting alone is insufficient.

**2. Preserve the owner's session when another request is released**

Adaptive serving admits multiple live requests, while the paired dFlash adapter keeps one speculative session. `TTScheduler._spec_session_owner` predicts which request can emit a full block. However, `Gemma4DFlashForCausalLM.release_request` ignores the released slot and clears the global speculative session, including a session owned by a different live request.

The failure sequence is:

1. `TTScheduler` schedules request A's solo prefill and commits A's anchor.
2. `TTScheduler` schedules request B's solo prefill and commits B's anchor. `Gemma4DFlashForCausalLM.prefill_forward` records B's pending session, and `TTScheduler._mirror_spec_session` records B as `_spec_session_owner`.
3. The client aborts A before B's first decode. `TTScheduler._mirror_spec_session` retains B because B remains in `self.requests`. `TTScheduler._update_after_schedule` reserves a full output block for B.
4. `TTModelRunner._update_states` processes A in `finished_req_ids` and calls `_release_model_request(A)`. `_release_model_request` calls `release_request(A_slot)`. See [model_runner.py:670-673](https://github.com/tenstorrent/vllm-tt-plugin/blob/7f13ca25abfde664e10870c47ef36ec00b170be1/src/vllm_tt_plugin/model_runner.py#L670-L673) and [model_runner.py:694-695](https://github.com/tenstorrent/vllm-tt-plugin/blob/7f13ca25abfde664e10870c47ef36ec00b170be1/src/vllm_tt_plugin/model_runner.py#L694-L695).
5. `Gemma4DFlashForCausalLM.release_request` clears B's `_spec_pending`, `_spec_pending_owner`, and `_spec_active` without checking `A_slot`. See [generator_vllm.py:3151-3165](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/generator_vllm.py#L3151-L3165).
6. `Gemma4DFlashForCausalLM.decode_forward` finds no session and routes B through one-token baseline decode. `TTScheduler._update_request_with_output` rejects B's one-token output because B's scheduling decision requires a full block. See [generator_vllm.py:3033-3043](https://github.com/tenstorrent/tt-metal/blob/ad2547245f0de5ca86f5f871bbd5c93aef21b798/models/demos/gemma4/tt/generator_vllm.py#L3033-L3043) and [scheduler.py:910-916](https://github.com/tenstorrent/vllm-tt-plugin/blob/7f13ca25abfde664e10870c47ef36ec00b170be1/src/vllm_tt_plugin/scheduler.py#L910-L916).

Requested change: Make the paired adapters' `release_request` behavior conditional on actual session ownership, and specify the same lifecycle contract for scheduler and model. Add a regression that prefills A and B separately, aborts A, applies runner lifecycle processing, and checks B's scheduled and emitted widths together.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -674,39 +739,202 @@ def reset_prefix_cache(
...
+        # output even when async scheduling has already run a later schedule().
+        if self._is_adaptive_block:
+            block_step = self._committing_block_step_decisions.get(request.request_id)
+            if block_step is None:
+                raise RuntimeError(
+                    "adaptive block-output request committed output without a "
+                    f"scheduling decision: req_id={request.request_id!r}"
+                )
+            if not block_step:
+                return super()._update_request_with_output(request, new_token_ids)
```
viktorpusTT, line 936: Please first test for `len(new_token_ids) != self.num_sampled_tokens_per_step` and raise a ValueError naming the request id, the expected width, the received width, and the fact that the scheduler stamped this step width 1.

viktorpusTT:
One more finding:

_spec_owner_slot holds a prefill-time state slot, while _release_model_request passes a post-remap decode row, so the new ownership check compares two quantities the plugin does not keep equal.
A concrete reachable sequence: `TTModelRunner._alloc_prefill_state_slots` gives request A slot 0. Request B then prefills solo at row 0, finds slot 0 held, and gets slot 1, so `Gemma4DFlashForCausalLM.prefill_forward` records `_spec_owner_slot = 1` for request B. The client aborts request A. Request B now decodes solo at row 0, so `TTModelRunner._decode_state_slot_remap` produces a non-identity permutation, issues the gather, sets `moved["req-b"] = 0`, and `TTModelRunner.note_decode_state_slots_settled` commits it. When request B finishes, `TTModelRunner._release_model_request` calls `release_request(0)` while `Gemma4DFlashForCausalLM` still holds `_spec_owner_slot == 1`, so the comparison at `models/demos/gemma4/tt/generator_vllm.py:3280` takes the early return at `:3287`. Request B's `_spec_pending`, `_spec_active`, `_spec_carry` and `_spec_owner_slot` all survive request B's death.
Suggested change: pick one identity for a request's model-side state and use it on both sides. Either have `Gemma4DFlashForCausalLM` consume `slot_remap` and move `_spec_owner_slot` with the gather, or have `TTModelRunner` release by the prefill slot it originally supplied.

arginugaTT:
> One more finding:
> 
> _spec_owner_slot holds a prefill-time state slot, while _release_model_request passes a post-remap decode row, so the new ownership check compares two quantities the plugin does not keep equal. A concrete reachable sequence: `TTModelRunner._alloc_prefill_state_slots` gives request A slot 0. Request B then prefills solo at row 0, finds slot 0 held, and gets slot 1, so `Gemma4DFlashForCausalLM.prefill_forward` records `_spec_owner_slot = 1` for request B. The client aborts request A. Request B now decodes solo at row 0, so `TTModelRunner._decode_state_slot_remap` produces a non-identity permutation, issues the gather, sets `moved["req-b"] = 0`, and `TTModelRunner.note_decode_state_slots_settled` commits it. When request B finishes, `TTModelRunner._release_model_request` calls `release_request(0)` while `Gemma4DFlashForCausalLM` still holds `_spec_owner_slot == 1`, so the comparison at `models/demos/gemma4/tt/generator_vllm.py:3280` takes the early return at `:3287`. Request B's `_spec_pending`, `_spec_active`, `_spec_carry` and `_spec_owner_slot` all survive request B's death. Suggested change: pick one identity for a request's model-side state and use it on both sides. Either have `Gemma4DFlashForCausalLM` consume `slot_remap` and move `_spec_owner_slot` with the gather, or have `TTModelRunner` release by the prefill slot it originally supplied.

Thanks Victor, I have pushed the fixes

src/vllm_tt_plugin/scheduler.py
arginugaTT: Done in 6c8561e. The baseline branch now checks `len(new_token_ids) != self.num_sampled_tokens_per_step` before delegating to `super()`, and raises a ValueError naming the request id, the received width, the expected width, and that the scheduler stamped this step width 1. Regression `test_a_block_width_output_on_a_baseline_step_is_refused` drives a real batched step to get width 1 stamped for both requests, then returns a full block for one of them; dropping the check fails it.

arginugaTT:
a passing vllm models test: https://github.com/tenstorrent/tt-metal/actions/runs/35375753037

viktorpusTT:
### 1. Cancelling one request can clear another request's speculative session

Commit 3a01d35 makes cleanup use the original prefill slot. That fixes the case where decode moves a request and cleanup subsequently fails to recognize its owner. However, another live request can reuse that original slot. Cleanup then mistakes the newer request for the older request:

1. `TTModelRunner._alloc_prefill_state_slots` places request X in slot 0 and request A in slot 1.
2. The client cancels X. `TTModelRunner._release_dead_state_slots` removes X's slot records.
3. `TTModelRunner._decode_state_slot_remap` moves A to slot 0. `note_decode_state_slots_settled` updates A's current slot, while `_req_prefill_state_slot` retains 1 for A.
4. The client submits B. `TTModelRunner._alloc_prefill_state_slots` assigns the now-free slot 1 to B. Gemma4's solo prefill arms B's speculative session with `_spec_owner_slot=1`.
5. The client cancels A before B decodes. `TTModelRunner._release_model_request` passes A's original slot 1. `Gemma4DFlashForCausalLM.release_request` treats the release as B's and clears B's session.
6. `TTScheduler._mirror_spec_session` retains live request B as owner. The next solo B decode has a block reservation, but Gemma4's no-session branch returns one token. `TTScheduler._update_request_with_output` rejects the width mismatch.

Change: must use a lifetime-unique request identity for ownership and release, or update model ownership through every physical slot remap and release by the corresponding current slot. Preserving the original slot alone is insufficient. Add a regression that includes slot movement, slot reuse, and cancellation of the older request while the newer owner survives.

### 2. Accepted small output blocks can exceed allocated KV space

Commit d51dae4 correctly adds KV reservation before execution. However, reserving 2 × output width does not cover every permitted verification width.

1. The operator selects `GEMMA4_DFLASH_SERVE_BLOCK=2` and `GEMMA4_DFLASH_VERIFY=7`. Gemma4 accepts output width 2 and uses eight physical verification rows. 
2. vLLM prefills a 123-token prompt using 128-token KV blocks and commits the prefill anchor.
3. `KVCacheManager.allocate_slots` reserves `123 + 1 + 4 = 128` token positions for the first decode. The real allocator retains only one request block, covering positions 0 through 127.
4. `DFlashFusedDecoder` supplies verification write positions 123 through 130. Positions 128 through 130 lack an allocated request block. 

This finding concerns permitted small output blocks; the reproduction does not establish insufficient allocation for the default 64-token output block.

Change: the plugin and model must agree on a bound for the complete physical KV extent, including verification and carry beyond emitted output. For a smaller immediate change, Gemma4 can reject output-width and verification-width combinations that violate a proven allocation bound.

arginugaTT:
@viktorpusTT , fixed your feedback in https://github.com/tenstorrent/vllm-tt-plugin/pull/118/changes/13ccab6016af160a0ec977f2cdaa6b6a75346efe

viktorpusTT (approved):

PR #119, opened 2026-09-10
[feat] Interleave decode steps between prefill chunks so long prefills do not starve decodes

### LINKED ISSUE #109: Interleave decode-only steps between prefill chunks so long chunked prefills do not starve decodes
## Summary

TT executes every step as all-prefill or all-decode; mixed prefill+decode batches are not supported. The scheduler's default policy is prefill-first: whenever any prefill work is pending (a waiting request, a `skipped_waiting` entry, or a running `is_prefill_chunk` continuation), the step is a prefill step, and decode runs only when the prefill pass cannot schedule a single token (`src/vllm_tt_plugin/scheduler.py:436-448`).

With scheduler-driven chunked prefill (#74), a long prompt is split across many steps, but under this policy those steps are consecutive: every running decode stalls for the entire duration of the multi-chunk prefill. That negates the main benefit chunked prefill was enabled for - bounding the head-of-line blocking a long prefill inflicts on other requests' ITL.

This issue proposes a scheduler-local heuristic that inserts decode-only steps between prefill-chunk steps, so running decodes keep advancing while a long prefill is in flight.

## Current behavior

`TTScheduler._schedule_without_pending_output_resumes` (`src/vllm_tt_plugin/scheduler.py:407-450`):

- `_has_pending_prefill()` (`:353-368`) is true when `waiting` or `skipped_waiting` is non-empty, or any running request has `is_prefill_chunk=True`.
- Default mode: pending prefill -> `_schedule_prefill_only()`. The decode-only fallback fires only when the prefill pass schedules zero tokens (KV pressure) and genuine decode requests exist (`:443`).
- Result: a 100K-token prompt at 8K tokens per chunk occupies ~13 consecutive steps, and no running request decodes in that window. The ITL spike duration scales with prompt length - exactly what chunked prefill was supposed to fix. This is already live for Gemma 4 (the one model with chunked prefill today) and becomes prominent once #74 lands.

Lane-DP has the same shape one level up: `TTLaneCoordinator._negotiate_forced_mode` (`src/vllm_tt_plugin/lane_scheduler.py:391-400`) forces a prefill step for all lanes whenever any lane reports prefill intent, and `_local_prefill_intent` (`:367-389`) always reports intent for a partial prefill. Standard multi-process DP ranks each run this policy independently; nothing cross-rank is needed.

Upstream vLLM does not have this problem: its token-budget scheduler mixes one prefill chunk and the running decodes into the same step. TT cannot mix, so the interleave granularity is a whole step and the scheduler must choose to alternate.

## Proposal

Add a decode-interleave policy to the default scheduling mode: when prefill work is pending AND running decode requests exist, allow a decode-only step once a configurable prefill-run limit is reached, instead of only when prefill makes zero progress.

Recommended starting heuristic (deliberately simple, deterministic, easy to test):

- Track consecutive prefill steps per scheduler. When the count reaches `K` (starting point: 4) and at least one genuine decode request is running, schedule one decode-only step and reset the counter.
- Hysteresis in both directions: after an interleaved decode step, require at least one prefill step before the next interleave, and cap consecutive decode steps while prefill is pending. The policy cannot oscillate every step, TTFT stays bounded, and prefill progress is guaranteed.

Alternatives worth weighing in the design discussion:

- Starvation-based: force a decode step when the oldest running decode has not advanced for N steps (or T wall-clock ms). Targets ITL directly, but needs per-request bookkeeping.
- Token-ratio: bound the prefill:decode token ratio over a sliding window. Closest to upstream's budget semantics, but harder to reason about because prefill and decode steps differ wildly in token count and latency.

Whichever heuristic is chosen, these constraints apply:

1. **Scheduler-local only.** The decision uses state the scheduler already has (queue contents, running set, step counters). No cross-rank coordination for standard DP; each rank interleaves independently.
2. **Lane-DP: the policy belongs in the coordinator**, not in per-lane schedulers. `_local_prefill_intent` / `_negotiate_forced_mode` must learn to pick decode when the coordinator's interleave state says so, otherwise lanes disagree on the forced mode. The coordinator-level zero-token fallback (`lane_scheduler.py:545-571`) stays unchanged.
3. **Keep the zero-progress fallback.** The existing KV-pressure fallback (prefill schedules nothing -> decode) is orthogonal and must keep working as is.
4. **Mode switches are not free.** Entering prefill drains pending async decode overlap (the steady-decode fast path; see `docs/SCHEDULING.md` and `docs/DECODE_RELOAD_CONTRACT.md`). Interleaving trades some decode-overlap efficiency for ITL smoothness, so the heuristic must not flip modes every step. Quantify this overhead in validation.
5. **No model-identity gating.** This is a scheduler policy; it applies whenever a chunked prefill is in flight, independent of model type, per the capability-gating contract.
6. **Configurable, conservative default.** Expose the aggressiveness (the prefill-run cap `K`, or an on/off switch) through engine args / additional config so operators can tune TTFT vs ITL. The default should change current behavior as little as possible while fixing the starvation case.

## Correctness notes

- A decode step cannot advance a partial prefill, and `_schedule_decode_only` (`scheduler.py:484-515`) already hides `is_prefill_chunk` requests. Interleaving changes only step ordering, not any per-request state machine. Greedy output must be identical to today.
- Fairness in both directions: decodes must not starve under a long prefill (the point of this issue), and the waiting / partial prefill must be guaranteed a step within a bounded number of decode steps (no TTFT regression beyond the configured cap).
- Block-output models never run chunked prefill (the platform disables it for them), so they are unaffected.

## Validation

- Host unit tests: drive `TTScheduler.schedule` with a scripted mix of one long chunked prefill plus running decodes; assert decode steps appear at the configured cadence and the prefill completes within the bounded window. Same for the lane coordinator's mode negotiation.
- Output equality: greedy decode with the policy on vs off, mixed long/short prompts.
- Device benchmarks (state host, model, mesh): ITL p50/p99 for concurrent decode requests while a long prompt prefills, TTFT for the long prompt, end-to-end throughput. Cover non-DP, lane-DP, and standard DP.

## Related

- #74 - enables scheduler-driven chunked prefill for the tt_transformers text models. This issue is the scheduling-policy follow-up that makes chunking deliver its ITL benefit.

viktorpusTT:
> [!NOTE]
> **Re-validated after the review fix.** Review found that `record_step` counted an ordinary decode step (one taken with no prefill pending) against the insertion allowance, so a newly arriving prompt could lose its first prefill step to an insertion it was never part of: `decode_interleave_decode_steps=D` delayed it by D-1 steps. D=1, the shipped default, was unaffected. Fixed, and the guarantee it contradicted is corrected in `SCHEDULING.md` and the `TTDecodeInterleavePolicy` docstring.
>
> Measured 2026-09-16 on `wh-lb-77` (T3K) against plugin `main` `3480a55` and tt-metal `5f29f785620`, which contains #55343 (`743890db`, required by plugin `main`'s #126 revert) as checked by `git merge-base --is-ancestor`. Submodules synced, built clean. Gated before testing on `_ttnncpp.so` existing, `import ttnn` succeeding, and the policy being deployed at default 2.
>
> | check | result |
> | --- | --- |
> | host unit tests | **103 passed** |
> | `tests/tt/test_chunked_prefill.py`, default 2/1 | **12 passed / 12** |
> | `tests/tt/test_chunked_prefill.py`, 2/2 (the only case the fix changes) | **6 passed / 6** |
> | `tests/tt/test_chunked_prefill.py`, policy off | **4 passed / 4** |
>
> Performance was not re-measured in this round; the tables below carry their own stack provenance.

## TL;DR

A TT step carries either prefill rows or decode rows, so a prompt the base scheduler splits into N chunks occupies N *consecutive* prefill steps and every running decode waits for all of them. This adds a scheduler-local policy that inserts decode-only steps into such a run. On a T3K the default setting cuts the worst decoder inter-token gap from 11152 ms to 899 ms for 1.0% on median time to first token.

### The problem

`TTScheduler._schedule_without_pending_output_resumes` prefers prefill whenever prefill work is pending. `TTScheduler._has_pending_prefill` stays true for the whole duration of a chunked prefill, because a partly prefilled request sits in `running` with `Request.is_prefill_chunk` set. So chunking a prompt N ways produces N consecutive prefill steps, and a concurrently decoding request advances in none of them.

Measurement on device shows this is not merely a missed optimization. It is a regression against not chunking at all. Counting the decoding request's gaps above three times its own baseline, as the arriving prompt grows and `max_num_batched_tokens` is 2048:

| arriving prompt (tokens) | 4096 | 8192 | 16384 | 24576 |
| --- | --- | --- | --- | --- |
| chunked prefill off | 1 | 1 | 1 | 1 |
| chunked prefill on, this policy off | **2** | **2** | **2** | **2** |
| chunked prefill on, this policy at 1 prefill step per insertion | 2 | 4 | 8 | 12 |

Splitting a prompt into 2, 4, 8 and 12 chunks gives the running decode **two** chances to advance, not 2/4/8/12. The two come from the pre-existing zero-progress fallback and the admission boundary, not from any policy. With this change the count is exactly the chunk count.

Upstream vLLM has no equivalent problem: its token-budget scheduler puts one prefill chunk and the running decodes in the same step. TT cannot mix, so the interleave granularity is a whole step and the scheduler has to choose to alternate.

### The change

`TTDecodeInterleavePolicy` (`src/vllm_tt_plugin/scheduler.py`) holds two integer counters and answers one question per step from state the scheduler already has:

- `wants_decode_step` returns true when the policy is enabled, prefill work is pending, a genuine decode is running, the consecutive-prefill-step count has reached `decode_interleave_prefill_steps`, and the decode count within the current insertion is below `decode_interleave_decode_steps`.
- `record_step` advances the counters with the phase the step actually ran.

`TTScheduler` consults it in DEFAULT mode only. `TTLaneCoordinator._negotiate_forced_mode` consults one coordinator-level instance and returns `TTSchedulingMode.DECODE_ONLY` in place of `TTSchedulingMode.PREFILL_ONLY`, because every lane executes the single negotiated mode and a per-lane decision would let lanes disagree. Standard multi-process DP needs nothing cross-rank: each rank runs the policy on its own scheduler state.

`TTPlatform._report_decode_interleave_policy` validates the keys at config time and logs the resolved policy unconditionally.

Four properties keep it contained:

1. It only ever chooses decode *in place of* a prefill step. It never chooses prefill, so it cannot delay a decode that would have run anyway.
2. It requires a running request a decode step can advance. A partial-prefill continuation does not qualify: `TTScheduler._schedule_decode_only` hides it and it samples no token.
3. The first step of any prefill run is never taken, so a request arriving while decodes run still gets its prefill step immediately.
4. An inserted step goes through `TTScheduler._schedule_decode_only`, the same pass every other decode step uses. Only step ordering changes.

The pre-existing zero-progress fallback (a prefill pass that schedules no tokens falls back to decode) is untouched.

### Configuration

Three keys under `additional_config['tt']`, read by `vllm_tt_plugin.config.get_tt_decode_interleave_config`:

| Key | Type | Default | Meaning |
| --- | --- | --- | --- |
| `decode_interleave_enabled` | bool | `true` | Off switch. When false the scheduler runs the strictly prefill-first policy that exists today. |
| `decode_interleave_prefill_steps` | int >= 1 | `2` | Consecutive prefill steps allowed before one decode-only step is inserted. |
| `decode_interleave_decode_steps` | int >= 1 | `1` | Decode-only steps that one insertion runs before a prefill step is required again. |

Both are **step** counts, not token counts, because the interleave granularity on TT is a whole step. Together they give a two-sided guarantee: a running decode advances at least once every `decode_interleave_prefill_steps + decode_interleave_decode_steps` steps, and pending prefill work gets a step at least once in the same window. Neither side starves and the policy cannot oscillate faster than that period.

```bash
python3 examples/server_example_tt.py \
  --model meta-llama/Llama-3.1-8B-Instruct \
  --max_num_seqs 32 --block_size 64 --max_model_len 32768 --async-scheduling \
  --enable-chunked-prefill --max-num-batched-tokens 2048 \
  --additional-config '{"tt": {"sample_on_device_mode": "all",
    "decode_interleave_prefill_steps": 2,
    "decode_interleave_decode_steps": 1}}'
```

## Related Artifacts

Resolves #109.

Follow-up to #74, which enables scheduler-driven chunked prefill for the tt_transformers text models. This PR is the scheduling-policy change that makes chunking deliver a bounded inter-token latency instead of a worse one.

Not a duplicate of any open PR. No open PR references #109, and no open PR touches the DEFAULT-mode prefill/decode phase choice. #118 (`arg/gemma4_spec_serving`) is the only other open PR that edits `src/vllm_tt_plugin/scheduler.py`; it changes block-output placeholder accounting in `_update_after_schedule` and `_update_request_with_output` and does not touch phase selection, so the overlap is a possible textual conflict in the import block, not a functional one.

No unmerged tt-metal dependency. tt-metal PRs 55630 and 55689 were needed for valid long-prompt measurements and are both merged; the branch is validated against stock tt-metal main.

## Notation used in the results

`K` and `D` below are shorthand for the arm labels only. They are **not** names in the code. An arm written `2/1` means `K=2, D=1`:

- `K` = `decode_interleave_prefill_steps`. A run of consecutive prefill steps is `K` steps long, so the gap a decode request sees spans `K` chunks. **`K` sets the inter-token-latency bound.**
- `D` = `decode_interleave_decode_steps`. The share of steps spent on prefill is `K / (K + D)`, so the **ratio** sets how device time splits between admitting new prompts and advancing running ones, which is what moves time to first token and throughput. `D` does not loosen the bound.
- `budget` = `max_num_batched_tokens`, the vLLM engine argument that sets the per-step token budget and therefore the chunk size. `vllm serve` and `examples/server_example_tt.py` default it to 2048, so the `budget 2048` rows are what an operator gets untuned.

Metric definitions, all in milliseconds unless stated:

- **ITL** = inter-token latency, the gap between two consecutive streamed tokens of one request. `ITL p50/p99/p99.9/max` are percentiles over the decoder gaps that fall inside the prefill window.
- **TTFT** = time to first token.
- **base ITL** = the median gap the streaming request produced *before* the long prompt was submitted, that is, its undisturbed cadence. It is 14.0 ms in every arm.
- **stalls** = count of decoder gaps above five times the decoder median gap (E2), or above three times the request's own base ITL (E3, E4).
- **stall total** = sum of those gaps' excess over the median, across all four decoders (E2).
- **excess wait** = sum of the streaming request's gaps above its own base ITL, that is, the entire cost of the long prefill as that request experienced it (E3).
- **tok/s** = output tokens per second over the whole run, all requests.

## Validation

https://github.com/tenstorrent/tt-metal/actions/runs/34844548474

Hardware: `wh-lb-77`, T3K (4x n300, 8 Wormhole chips), `MESH_DEVICE=T3K`. Model `meta-llama/Llama-3.1-8B-Instruct`. vLLM 0.26.0.

**Two stacks are involved, and the tables say which.** The performance tables in the sections below were measured on tt-metal `398bd7a5ec2` with PRs 55630 and 55689 as working-tree patches, before those merged. Correctness and the decision-critical performance arms were then re-measured on tt-metal `e3ca937f19c` (stock main, 204 commits later, submodules synced and rebuilt) with plugin `main` `52db643`. The cross-stack agreement is in "Re-measured on the current stack" below: every arm reproduces within 0.7%, so the older tables stand.

Server flags common to every arm: `--max_num_seqs 32 --block_size 64 --max_model_len 32768 --async-scheduling --no-enable-prefix-caching`, with `fabric_config: FABRIC_1D`, `sample_on_device_mode: all`, `trace_region_size: 85000000`. Prefix caching is off so a repeated prompt is prefilled again. Async scheduling is on, which is the configuration in which a mode switch costs the most, so the overhead figures are the pessimistic ones.

`TTScheduler` in DEFAULT mode throughout: single-process non-DP, one engine over the whole 8-chip mesh.

### Re-measured on the current stack

The branch as it now stands: tt-metal `e3ca937f19c` (stock main), plugin `main` `52db643`, no working-tree patches anywhere. Deployment was verified before measuring (`TTDecodeInterleavePolicy` present, `#122` warmup fix present, resolved default 2, tt-metal clean).

Correctness, `tests/tt/test_chunked_prefill.py`, the CI invocation repeated against one server per arm:

| arm | result |
| --- | --- |
| interleave on, default 2/1 | **12 passed / 12** |
| interleave off | 4 passed / 4 |

Performance, E2 mixed load, median of three reps, old stack against new:

| arm | worst gap old | new | delta | median TTFT old | new | delta |
| --- | --- | --- | --- | --- | --- | --- |
| off | 11148.0 ms | 11162.8 ms | +0.1% | 4330.3 ms | 4340.2 ms | +0.2% |
| 4/1 | 1603.7 ms | 1604.9 ms | +0.1% | 4229.9 ms | 4259.9 ms | +0.7% |
| 2/1 (default) | 901.8 ms | 900.6 ms | -0.1% | 4388.0 ms | 4381.9 ms | -0.1% |
| 2/2 | 531.9 ms | 533.2 ms | +0.2% | 4586.3 ms | 4601.4 ms | +0.3% |
| 1/1 | 524.1 ms | 522.0 ms | -0.4% | 4675.5 ms | 4671.5 ms | -0.1% |

204 tt-metal commits and the #122 warmup change moved every figure by less than one percent, so the conclusions drawn from the older tables are unaffected.

The same five arms, stated entirely on the current stack and relative to the interleave-off arm measured beside them:

| arm | worst decode gap | vs off | median TTFT | throughput |
| --- | --- | --- | --- | --- |
| off | 11162.8 ms | 1.0x | +0.0% | +0.0% |
| 4/1 | 1604.9 ms | 7.0x | -1.9% | -0.7% |
| **2/1 (default)** | **900.6 ms** | **12.4x** | **+1.0%** | **-1.3%** |
| 2/2 | 533.2 ms | 20.9x | +6.0% | -2.0% |
| 1/1 | 522.0 ms | 21.4x | +7.6% | -2.5% |

E1, E3, E4 and the async-mechanism table were **not** re-measured on the current stack; they are the old-stack figures. Given the agreement above that is a judgement call, not a verified claim, and a reviewer who wants them re-run should say so.

### Host unit tests

```text
$ .venv/bin/python -m pytest tests/test_decode_interleave.py tests/test_scheduler.py \
    tests/test_lane_scheduler.py tests/test_chunked_prefill_policy.py -q
72 passed
```

Same 72 pass in the tt-metal environment on `wh-lb-77`.

No new host-suite failures. The comparison ran on a machine without ttnn, where some tests fail for that reason alone, so the baseline was taken from a separate `git worktree` at `origin/main` (`c9cfebc`) with that worktree's own `src` first on `PYTHONPATH`, which shadows the editable install and guarantees the baseline executes `c9cfebc` code. Over the fourteen host test files that collect without ttnn, `c9cfebc` fails 15 tests and this branch fails the same 15 by name, all with `ModuleNotFoundError: No module named 'ttnn'`.

```text
$ pre-commit run --all-files
all hooks passed
```

### E1: time to first token with nothing to interleave (control)

One 16384-token prompt at concurrency 1, nothing else running, so `has_running_decode` is false and the policy cannot fire. One warmup plus three measured runs.

| arm | measured runs | median |
| --- | --- | --- |
| chunked prefill off | 1870.8, 1867.9, 1875.0 | 1870.8 |
| budget 2048, interleave off | 2888.6, 2894.0, 2889.3 | 2889.3 |
| budget 2048, K=1 D=1 | 2882.5, 2898.5, 2888.7 | 2888.7 |
| budget 2048, K=2 D=1 | 2883.4, 2885.6, 2882.1 | 2883.4 |
| budget 2048, K=4 D=1 | 2888.3, 2867.2, 2890.0 | 2888.3 |
| budget 8192, interleave off | 1896.3, 1899.1, 1904.4 | 1899.1 |
| budget 8192, K=1 D=1 | 1900.5, 1901.5, 1899.5 | 1900.5 |

The four `budget 2048` arms agree within 6 ms (0.2%), the two `budget 8192` arms within 1.4 ms. The policy is inert when there is nothing to interleave.

This table also prices chunked prefill itself, which this PR does not change: splitting one 16384-token prompt eight ways at `budget 2048` raises its TTFT from 1870.8 to 2889.3 ms (+54%); splitting it two ways at `budget 8192` raises it to 1899.1 ms (+1.5%).

### E2: mixed serving load

Four decoder requests (512-token prompt, 768 streamed output tokens) start at t=0. Three seconds later four prefiller requests (16384-token prompt, 32 output tokens) are submitted two seconds apart. Median of three reps.

| arm | ITL p50 | ITL p99 | ITL p99.9 | ITL max | stalls | stall total | TTFT p50 | TTFT max | tok/s |
| --- | --- | --- | --- | --- | --- | --- | --- | --- | --- |
| chunked prefill off | 14.3 | 44.2 | 1853.0 | 1853.1 | 16 | 29322.7 | 1886.7 | 1890.8 | 164.6 |
| budget 2048, interleave off | 14.3 | 27.6 | 11150.6 | **11152.5** | 8 | 45448.0 | 4338.0 | 5366.4 | 138.1 |
| budget 2048, K=4 D=1 | 14.3 | 1297.1 | 1603.3 | 1603.4 | 36 | 45894.0 | 4228.0 | 5578.3 | 137.3 |
| budget 2048, K=2 D=1 (default) | 14.3 | 742.5 | 898.6 | **898.6** | 68 | 46473.4 | 4380.7 | 5838.8 | 136.4 |
| budget 2048, defaults, no key set | 14.4 | 744.4 | 899.2 | 898.0 | 68 | 46492.1 | 4382.5 | 5841.5 | 136.3 |
| budget 2048, K=1 D=1 | 14.3 | 481.8 | 522.8 | 523.1 | 128 | 47483.4 | 4660.0 | 6348.0 | 131.5 |
| budget 8192, interleave off | 14.3 | 854.1 | 1074.1 | 1074.2 | 32 | 30248.8 | 1913.5 | 1914.7 | 164.0 |
| budget 8192, K=1 D=1 | 14.3 | 853.8 | 1062.1 | 1062.2 | 32 | 30033.6 | 1948.1 | 1956.4 | 163.6 |

Per-rep spread: `ITL max` varies by at most 4 ms across an arm's three reps and `TTFT p50` by at most 20 ms. Output throughput is the noisiest column (the `K=1 D=1` arm produced 131.5, 134.7 and 122.9 tok/s), so read it with a spread of several tok/s and the rest as exact.

Four readings:

- **Enabling chunked prefill at `budget 2048` makes the worst decoder gap six times worse than not splitting prompts at all**: 1853.1 to 11152.5 ms. The `stalls` column shows why. Chunked prefill off gives 16 stalls, one per prefiller per decoder. Chunked prefill on with the interleave off gives 8, each far longer: four prefillers of eight chunks are 32 prefill steps, and because a prefiller arrives before the previous one finishes, `_has_pending_prefill` never goes false across them.
- **The interleave removes that regression and beats the unsplit baseline.** At the default the worst gap is 898.6 ms: 12.4x better than the same budget with the policy off, and 2.1x better than chunked prefill off.
- **The shipped default reaches the scheduler.** The "defaults, no key set" row is a server started with no `decode_interleave_*` key. Its log line reads `Decode interleave enabled: at most 2 consecutive prefill step(s) before 1 decode-only step(s)`, and it matches the explicit `K=2 D=1` arm to within 2 ms on both worst gap and median TTFT.
- **At `budget 8192` the policy changes almost nothing** (1074.2 to 1062.2 ms, 1.1%), because that arrival pattern never queues more than two chunks. Note also that the `budget 8192` arms beat every `budget 2048` arm on worst gap, TTFT and throughput. This policy rescues a small token budget; it does not make a small budget preferable to a large one.

### E3: one streaming decode against one 24576-token prompt

Request A streams 1024 tokens from a 512-token prompt. Once A has streamed 32 tokens, request B fires with a 24576-token prompt and `max_tokens=1`. Median of five reps.

| arm | A base ITL | A worst gap | A excess wait | A stalls | B TTFT |
| --- | --- | --- | --- | --- | --- |
| chunked prefill off | 14.0 | 4443.8 | 4791.0 | 1 | 4474.9 |
| budget 2048, interleave off | 14.0 | 4956.6 | 5517.9 | 2 | 5202.3 |
| budget 2048, K=4 D=1 | 14.0 | 1890.7 | 5521.7 | 4 | 5243.1 |
| budget 2048, K=2 D=1 (default) | 14.0 | 1181.5 | 5537.8 | 7 | 5296.2 |
| budget 2048, defaults, no key set | 14.0 | 1180.7 | 5555.8 | 7 | 5309.3 |
| budget 2048, K=1 D=1 | 14.0 | 657.8 | 5572.9 | 12 | 5418.8 |
| budget 8192, interleave off | 14.0 | 2273.9 | 3426.5 | 2 | 3123.2 |
| budget 8192, K=1 D=1 | 14.0 | 1245.4 | 3445.1 | 3 | 3165.8 |

The `excess wait` column is the honest framing. A waits essentially the same total time in every `budget 2048` arm, 5518 to 5573 ms, a spread of 1.0%. The prefill work is unchanged, so A's total wait cannot shrink. What changes is the shape: one gap of 4957 ms becomes twelve gaps of at most 658 ms. A request streaming to a user perceives the worst gap, not the sum. That 1.0% growth in excess wait is the mode-switch overhead, and it is the same effect that raises B's TTFT by 4.2%.

### E4: the bound does not grow with the arriving prompt's length

E3 repeated at four B prompt lengths. Median of three reps.

Worst gap the streaming request saw (ms):

| arm | 4096 | 8192 | 16384 | 24576 | growth |
| --- | --- | --- | --- | --- | --- |
| chunked prefill off | 430.4 | 839.9 | 1838.7 | 4431.9 | 10.3x |
| budget 2048, interleave off | 281.0 | 910.2 | 2627.4 | 4953.2 | 17.6x |
| budget 2048, K=1 D=1 | 278.2 | 353.8 | 504.5 | 658.6 | **2.4x** |
| budget 2048, K=1 D=2 | 267.9 | 341.9 | 495.1 | 646.0 | 2.4x |

B's TTFT over the same runs (ms):

| arm | 4096 | 8192 | 16384 | 24576 |
| --- | --- | --- | --- | --- |
| chunked prefill off | 453.7 | 864.6 | 1865.7 | 4464.5 |
| budget 2048, interleave off | 520.3 | 1149.2 | 2868.7 | 5197.6 |
| budget 2048, K=1 D=1 | 535.9 | 1209.5 | 3020.3 | 5420.3 |
| budget 2048, K=1 D=2 | 551.2 | 1251.0 | 3110.5 | 5568.5 |

Count of the streaming request's gaps above three times its base ITL, which is the mechanism stated directly:

| arm | 4096 | 8192 | 16384 | 24576 |
| --- | --- | --- | --- | --- |
| chunked prefill off | 1 | 1 | 1 | 1 |
| budget 2048, interleave off | 2 | 2 | 2 | 2 |
| budget 2048, K=1 D=1 | 2 | 4 | 8 | 12 |
| budget 2048, K=1 D=2 | 2 | 4 | 8 | 12 |

With the interleave on the count is exactly B's chunk count (4096/2048, 8192/2048, 16384/2048, 24576/2048): every chunk boundary becomes a point at which the decode advances.

The worst gap at `K=1` still grows 2.4x while the prompt grows 6x. It grows because the bound is one chunk and a chunk costs more deeper into a prompt: the chunked SDPA attends over every token cached so far, so the twelfth chunk costs more than the first. The bound holds; the unit of the bound is not constant.

`D=2` is not worth its cost: it improves the worst gap by 2% to 4% and raises B's TTFT by 2.7% at 24576 tokens. `decode_interleave_decode_steps` stays at 1.

### Why the default is 2

From the E2 table, against `budget 2048, interleave off`:

| `decode_interleave_prefill_steps` | worst decoder gap | vs off | median TTFT | vs off | throughput | vs off |
| --- | --- | --- | --- | --- | --- | --- |
| policy off | 11152.5 | 1.0x | 4338.0 | 0% | 138.1 | 0% |
| 4 | 1603.4 | 7.0x | 4228.0 | -2.5% | 137.3 | -0.6% |
| **2 (default)** | **898.6** | **12.4x** | **4380.7** | **+1.0%** | **136.4** | **-1.2%** |
| 1 | 523.1 | 21.3x | 4660.0 | +7.4% | 131.5 | -4.8% |

2 is the knee. Going 4 to 2 buys a further 1.8x on the worst gap for 3.5 points of TTFT; going 2 to 1 buys a further 1.7x for 6.4 more points of TTFT and 3.6 more points of throughput.

Issue #109 suggests 4 as a starting point. The measurements do not support 4 as the default: it leaves a 1603 ms worst-case gap, still 112x the 14.3 ms baseline ITL.

The `K=4` improvement in TTFT is consistent across all three reps (4213.0, 4228.0, 4236.0 against 4338.0, 4328.4, 4338.5), so it is not noise, but the mechanism behind it is not verified here.

The `D` sweep below supersedes what an earlier revision of this description said about `K=1` being the setting for ITL priority. It is not; `2/2` reaches the same bound more cheaply.

### Sweeping `K` against `D`

The tables above hold `D` at 1 except for one arm, which left the knob where mode-switch cost lives almost untested. Ten settings at budget 2048, same conditions. Median of three reps for the mixed-load columns, four for the prefill-window columns. The last two columns come from the single-decoder experiment, whose 24576-token prompt splits into twelve chunks.

| K / D | worst decode gap | median TTFT | vs off | throughput | vs off | decode tokens during a 12-chunk prefill | added TTFT per extra decode token |
| --- | --- | --- | --- | --- | --- | --- | --- |
| off | 11,148 ms | 4,330 ms | +0.0% | 138.1 tok/s | +0.0% | 2 | n/a |
| 8 / 4 | 2,657 ms | 4,289 ms | -1.0% | 136.9 tok/s | -0.9% | 6 | 20.8 ms |
| 4 / 1 | 1,604 ms | 4,230 ms | -2.3% | 137.2 tok/s | -0.7% | 4 | 27.9 ms |
| 4 / 2 | 1,393 ms | 4,335 ms | +0.1% | 136.8 tok/s | -0.9% | 6 | 21.1 ms |
| 4 / 4 | 1,396 ms | 4,540 ms | +4.9% | 135.7 tok/s | -1.7% | 10 | 16.6 ms |
| 2 / 1 **(default)** | 902 ms | 4,388 ms | +1.3% | 136.2 tok/s | -1.4% | 7 | 23.1 ms |
| 2 / 2 | 532 ms | 4,586 ms | +5.9% | 132.7 tok/s | -3.9% | 12 | 19.2 ms |
| 2 / 4 | 531 ms | 5,006 ms | +15.6% | 133.7 tok/s | -3.2% | 22 | 16.1 ms |
| 1 / 1 | 524 ms | 4,676 ms | +8.0% | 134.6 tok/s | -2.5% | 13 | 21.2 ms |
| 1 / 2 | 510 ms | 5,106 ms | +17.9% | 132.8 tok/s | -3.8% | 24 | 17.5 ms |

The interleave-off arm was re-measured on a fresh reservation container a day after the earlier tables and returned 11,148 ms and 4,330 ms against 11,152 ms and 4,338 ms, so the two sets are comparable.

Three results:

**Turning the policy off is beaten on both measured axes, but not because the policy does less work.** `4/1` and `8/4` both show a lower worst decode gap *and* a lower median TTFT than off. The second half is a token-publish effect, not a work reduction: the interleave strictly adds steps, so prefill finishes strictly later. Under async scheduling a step's token is published only after the *next* step is submitted, and on TT a prefill step is a whole synchronous chunk, so with the interleave off a prompt's first token waits a full chunk for the next queued prompt's chunk, while with it on that next step is usually a short decode step. Turning async scheduling off reverses the comparison, per-prompt TTFT, median of three reps:

| arm | async | prompt 1 | 2 | 3 | 4 | median | worst gap |
| --- | --- | --- | --- | --- | --- | --- | --- |
| off | on | 3090 | 3916 | 4745 | 5361 | 4330 ms | 11,148 ms |
| 4 / 1 | on | 2892 | 3788 | 4682 | 5589 | 4230 ms | 1604 ms |
| off | off | 2848 | 3673 | 4491 | 5317 | 4082 ms | 11,340 ms |
| 4 / 1 | off | 2871 | 3773 | 4671 | 5575 | 4223 ms | 1759 ms |

The off arm gains 242, 243, 254 and 44 ms from disabling async scheduling; `4/1` gains 21, 15, 11 and 14 ms. Without async scheduling the interleave costs **+3.5%** on median TTFT and every prompt is slower, which is the honest device-work price. The worst-gap improvement is not an artifact: it survives async-off at 11,340 against 1759 ms. So read the TTFT column as "under async scheduling a conservative setting costs nothing client-visible", not as "interleaving is free".

**`D` is a real knob, and a cheaper one than `K`.** The price of one interleaved decode token falls from 27.9 ms at `4/1` to 16.1 ms at `2/4`, monotonically with `D` at every `K`. That is the fixed cost of leaving and re-entering decode being spread over a longer decode run.

**`1/1` should not be used.** `2/2` reaches the same bound, 532 ms against 524 ms, a difference inside the 4 ms run-to-run spread of that metric, for 5.9% on TTFT rather than 8.0%. Reaching a tight bound by shortening the prefill run costs more than reaching it by lengthening the decode run.

Settings worth knowing:

| goal | setting | worst decode gap | median TTFT, async on | median TTFT, async off |
| --- | --- | --- | --- | --- |
| TTFT first | `4/1` | 1604 ms | -2.3% | +3.5% |
| balanced, shipped default | `2/1` | 902 ms | +1.3% | not measured |
| ITL first | `2/2` | 532 ms | +5.9% | not measured |

`2/4`, `1/1`, `1/2`, `4/4` and `8/4` are each beaten on both axes by one of those three at this budget and workload.

### Why `2/2` halves the gap, which `K` alone does not explain

`2/2` reaching 532 ms contradicts "a run is `K` steps long, so the gap spans `K` chunks". Two prefill steps should give a two-chunk gap, which is what `2/1` delivers. The gap arithmetic shows what happens: at `2/2` the twelve gaps come in pairs whose sums match `2/1`'s two-chunk gaps (283 + 307 = 590 against 576; 360 + 383 = 743 against 731), so one of each run's two tokens arrives *inside* the next prefill run.

Lining the sequences up shows the interval between two decode tokens is unchanged to within 1.0-2.5%, and the pair splits it almost exactly in half:

| `2/1` gap | matching `2/2` pair | pair sum | vs `2/1` | split |
| --- | --- | --- | --- | --- |
| 576.0 ms | 283.4 + 307.3 | 590.7 ms | 102.5% | 48 / 52 |
| 731.5 ms | 359.5 + 383.1 | 742.6 ms | 101.5% | 48 / 52 |
| 884.4 ms | 435.1 + 458.8 | 893.9 ms | 101.1% | 49 / 51 |
| 1028.0 ms | 509.3 + 534.8 | 1044.1 ms | 101.6% | 49 / 51 |
| 1182.6 ms | 585.4 + 609.4 | 1194.8 ms | 101.0% | 49 / 51 |

What moves is *when the run's second token is published*, not how much work ran; the residue is the extra decode step. Cause: `EngineCore.step_with_batch_queue` submits step N and returns without touching any output while the batch queue has room (`vllm/v1/engine/core.py:673`), then pops the oldest entry and calls `update_from_output` on it (`:703`). With async scheduling `max_concurrent_batches` is 2 (`vllm/config/vllm.py:512`), so step N-1's tokens are published only after step N is submitted, and on TT a final prefill is sampled synchronously (`src/vllm_tt_plugin/model_runner.py:2567`), so that submission costs a whole chunk. In a `P P D D` cycle the first decode token is followed by a decode step and the second by a prefill chunk, which is the staggering measured above. Tested, not inferred, by re-running both with `--no-async-scheduling`:

| setting | async scheduling | worst gap | insertions | decode tokens | tokens per insertion | steady ITL |
| --- | --- | --- | --- | --- | --- | --- |
| 2 / 1 | on | 1183 ms | 7 | 7 | 1.00 | 13.98 ms |
| 2 / 2 | on | 662 ms | 12 | 12 | 1.00 | 13.98 ms |
| 2 / 1 | off | 1255 ms | 6 | 6 | 1.00 | 15.94 ms |
| 2 / 2 | off | 1256 ms | 6 | 11 | 1.83 | 16.08 ms |

Two things follow.

1. **The state machine does what its docstring says.** With async off, `2/2` delivers 1.83 tokens per insertion over 6 insertions, so the cadence really is two prefill steps then two decode steps. The `1.00` in the async-on rows is the same run with one of its two tokens released later, not a shorter run.
2. **The extra halving at `D >= 2` is an async-overlap effect, not a scheduling effect,** so it is available only with async scheduling on. That is the default for the tt_transformers text models. A block-output model must run `--no-async-scheduling` and would not see it, but block-output models never run chunked prefill, so the combination does not arise.

The same table prices async decode overlap itself, which this PR does not touch: steady inter-token latency is 13.98 ms with it and 15.94 ms without.

**Should the publish delay be changed?** No. The one-step delay is inherent to vLLM async scheduling; it *is* the overlap mechanism, and the synchronous alternative is exactly what `--no-async-scheduling` selects. What is TT-specific is how long the delay lasts, since a TT prefill step is a whole synchronous chunk. Publishing the pending decode output before submitting the next chunk is the intervention that would restore the clustered pattern, and the async-off arms measure what that costs: `2/2` goes from a 662 ms worst gap to 1256 ms. The current behavior spends nothing to halve the worst gap, so it should be left alone, with the caveat that the `2/2` recommendation depends on async scheduling being on. A block-output model runs `--no-async-scheduling` and would not get the halving, but block-output models never run chunked prefill.

### Output correctness

The intended claim was byte-identical greedy output with the policy on and off. **That claim cannot be established by bit-exact comparison on this device, for a reason unrelated to this PR.** Capturing the same 13 greedy prompts twice on one server, `temperature=0.0`, `seed=0`:

| comparison | result |
| --- | --- |
| interleave off, pass a vs pass b (same server) | differs on 2 of 13 |
| K=1 D=1, pass a vs pass b (same server) | differs on 3 of 13 |
| interleave off pass a vs K=1 pass a | differs on 2 of 13 |
| interleave off pass b vs K=1 pass b | differs on 2 of 13 |

The device does not reproduce its own text within one process, and the cross-arm difference rate is no larger than the same-arm control. The decisive control is a 1024-token prompt sent alone at `budget 2048`: it is under the token budget so it is not chunked, and nothing else is running, so the policy provably cannot fire on it, yet one arm's text differed there while two others matched the baseline. A difference on a request the policy cannot reach is device nondeterminism. This is consistent with `TTSampling._adjust_values_for_tiebreak` unifying only exactly-equal maxima, so a 1-ULP logit difference flips a token when the argmax margin is tight.

The check that does hold is semantic. Every probe prompt hides a passphrase at 60% depth and ends by asking for it back, which survives a token flip that only rewords the answer. Over all eleven captures:

```text
budget 2048, K=1 D=1   (sweep)              13/13
budget 2048, K=2 D=1   (sweep)              13/13
budget 2048, K=4 D=1   (sweep)              13/13
budget 8192, K=1 D=1   (sweep)              13/13
budget 2048, K=1 D=1   (correctness pass a) 13/13
budget 2048, K=1 D=1   (correctness pass b) 13/13
chunked prefill off    (sweep)              13/13
budget 2048, off       (sweep)              13/13
budget 8192, off       (sweep)              13/13
budget 2048, off       (correctness pass a) 13/13
budget 2048, off       (correctness pass b) 12/13   <-- only miss, interleave OFF
```

The single miss in the whole dataset is in an interleave-off arm: it emitted `" eq\nThe NATO phonetic alphabet is a set of letters that corresponds to the first 26 letter"`, having started the passphrase `eq-short-1` and diverged after the token `eq`. Both `K=1 D=1` passes produced it correctly.

What actually carries the correctness claim is structural: an inserted step calls `TTScheduler._schedule_decode_only`, the identical pass every other decode step uses, which hides `waiting`, hides `skipped_waiting`, removes the `is_prefill_chunk` entries from `running`, and restores all three. `TTDecodeInterleavePolicy` writes nothing except its own two integers and touches no `Request`, no `SchedulerOutput` and no KV state. `tests/test_decode_interleave.py::test_interleaved_decode_step_hides_prefill_work` asserts the hiding and restoration.

### Not covered on device

- **Single-process lane-DP.** Lane mode requires the Galaxy DP-to-lanes conversion and `wh-lb-77` is a T3K, so it cannot run there. Host coverage is `tests/test_lane_scheduler.py::test_decode_interleave_overrides_a_prefill_intent_once_the_bound_is_hit`, `::test_decode_interleave_needs_a_lane_with_a_genuine_running_decode` and `::test_schedule_records_the_step_phase_for_the_interleave_policy`.
- **Standard multi-process DP.** Each rank runs its own `TTScheduler` in DEFAULT mode with its own policy instance, which is the code path E1 to E4 exercised, with no cross-rank coordination. `TTScheduler` ignores upstream's `throttle_prefills` argument both before and after this change.
- One model, one mesh. The ratio between a prefill chunk and a decode step sets the whole tradeoff and differs elsewhere, so an operator on different hardware should re-run E2 before trusting the default.
- The E2 two-second arrival spacing is a chosen parameter. Wider spacing shortens a run of prefill steps and shrinks the effect.
- Nothing here measures a prompt near the 32768 context limit; E4 stops at 24576.

## Checklist

---

AI assistance was used for the implementation, the measurement harness, the analysis and this description. The submitting human must review every changed line and validate the change before merge.


docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+has to choose to alternate.
+
+`TTDecodeInterleavePolicy` (in `scheduler.py`) bounds that run. It holds two
+counters and answers one question per step, from state the scheduler already
+has:
+
+- `decode_interleave_prefill_steps` (default 2): consecutive prefill steps
+  allowed before a decode-only step is inserted.
+- `decode_interleave_decode_steps` (default 1): decode-only steps that
+  insertion runs before a prefill step is required again.
```
tchedaTT: before a prefill step is required, or before a prefill step is allowed again?

docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+advances at least once every `prefill_steps + decode_steps` steps, and pending
+prefill work gets a step at least once every `prefill_steps + decode_steps`
+steps. Neither side starves, and the policy cannot oscillate faster than that
+period.
+
+Four properties keep the policy contained:
+
+- It only ever chooses decode *in place of* prefill. It never chooses prefill,
+  so it cannot delay a decode that would have run anyway.
+- It requires a running request that a decode step can actually advance. A
```
tchedaTT: I'm not sure what this is trying to say

docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+  partial-prefill continuation does not count, because a decode step samples no
+  token for one.
+- The first step of any prefill run is never taken. A request arriving while
```
tchedaTT: I'm not sure this is true - see the scenario codex came up with

docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+One effect is worth knowing because it is not a scheduling effect at all. With
+async scheduling on, a decode step's output is applied one step later and
+entering prefill drains it, so the last decode output of an insertion is
+released only when the next prefill step runs. That splits a `prefill_steps`
+chunk gap into one chunk plus the rest, which is why 2 prefill steps with 2
+decode steps reaches a 532 ms worst gap where 2 prefill steps with 1 decode
+step reaches 902 ms. Re-running both with `--no-async-scheduling` returns them
+to the same 1255 ms two-chunk gap, confirming the cause. A block-output model
+runs with `--no-async-scheduling` and would not see that halving, but such a
+model never runs chunked prefill anyway.
```
tchedaTT: I'm not sure what this is trying to say, this needs rewriting/clarification. I get the draining part, but what is a chunk gap? What does it mean by "the last decode output of an insertion is
released only when the next prefill step runs"? In what sense is it released?

docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+more than reaching it by lengthening the decode run.
+
+In single-process lane-DP the policy lives in `TTLaneCoordinator`, not in the
+per-lane schedulers. Every lane executes the one negotiated mode, so a per-lane
+decision would let lanes disagree about it. `_negotiate_forced_mode` consults
+one coordinator-level policy and overrides a prefill intent with
+`DECODE_ONLY` when the bound is reached and some lane holds a genuine running
+decode. Standard multi-process DP needs nothing cross-rank: each rank runs the
+policy on its own scheduler state.
+
```
tchedaTT: Please simplify this and get rid of the agentspeak. There are easier ways to say that for lane-dp we insert he decodes if some lane has a decode that can be inserted.

docs/SCHEDULING.md
```diff
@@ -190,6 +195,136 @@ Two things follow for the TT execution path:
...
+| --- | --- | --- | --- |
+| time to first token first | 4 / 1 | 1604 ms | -2.3% |
+| balanced, the default | 2 / 1 | 902 ms | +1.3% |
+| inter-token latency first | 2 / 2 | 532 ms | +5.9% |
+
+Switching the policy off is not among them: 4 / 1 gives both a lower worst gap
+and a lower time to first token than off. Nor is 1 / 1, which reaches the same
+bound as 2 / 2 (524 ms against 532, inside the run-to-run spread) for 8.0%
+rather than 5.9%: reaching a tight bound by shortening the prefill run costs
+more than reaching it by lengthening the decode run.
```
tchedaTT: Do we need this whole chunk here?

src/vllm_tt_plugin/config.py
```diff
@@ -179,3 +179,69 @@ def uses_tt_lane_coordinator(vllm_config: "VllmConfig") -> bool:
...
+_DECODE_INTERLEAVE_ENABLED_KEY = "decode_interleave_enabled"
+_DECODE_INTERLEAVE_PREFILL_STEPS_KEY = "decode_interleave_prefill_steps"
+_DECODE_INTERLEAVE_DECODE_STEPS_KEY = "decode_interleave_decode_steps"
+
+_DECODE_INTERLEAVE_ENABLED_DEFAULT = True
+# Measured on a T3K with Llama-3.1-8B-Instruct at max_num_batched_tokens=2048,
+# four 16384-token prompts arriving against four streaming decodes: 2 bounds
+# the worst decode gap at 899 ms against 11152 ms with the policy off, for 1.0%
+# on median time to first token. 1 reaches 523 ms but costs 7.4%; 4 leaves
+# 1603 ms.
```
tchedaTT: Does this belong here?

src/vllm_tt_plugin/lane_scheduler.py
```diff
@@ -388,15 +393,35 @@ def _local_prefill_intent(self, sched: TTScheduler) -> int:
...
 
         The device executes all lanes together and cannot mix prefill with
         decode, so the lanes must agree. If *any* lane wants to prefill, the
         whole step is prefill-only; otherwise it is decode-only. Lanes without
         work for the chosen mode simply contribute an empty batch.
+
+        The decode-interleave policy overrides a prefill intent once a run of
+        prefill steps reaches its bound, so the decision stays a single
+        coordinator-level choice: taking it per lane would let lanes disagree
+        about the shared mode.
```
tchedaTT: Just say sth like "Decode interleaving is decided jointly for all lanes"

src/vllm_tt_plugin/platform.py
```diff
@@ -172,6 +173,31 @@ def _disable_chunked_prefill(vllm_config: "VllmConfig", reason: str) -> None:
...
 
+def _report_decode_interleave_policy(vllm_config: "VllmConfig") -> None:
+    """Validate the decode-interleave settings and record the resolved policy.
+
+    Called for its raising side effect as much as for the log line: an invalid
+    ``decode_interleave_*`` value must stop the server at config time rather
+    than at the first scheduler step in a worker subprocess. The scheduler
+    config is absent from ``/metrics`` and an interleaved decode step is
+    indistinguishable from an ordinary one in the iteration stats, so this line
+    is the only external signal of the active policy.
```
tchedaTT: verbose

src/vllm_tt_plugin/scheduler.py
```diff
@@ -55,6 +59,87 @@ def from_prefill_intent(cls, prefill_intent: int) -> "TTSchedulingMode":
...
+    chunks are simply consecutive.
+
+    This policy inserts decode-only steps into such a run. It holds two
+    counters and answers one question per step, from state the scheduler
+    already has:
+
+    - ``prefill_steps``: how many consecutive prefill steps run before a
+      decode-only step is inserted.
+    - ``decode_steps``: how many decode-only steps that insertion runs before a
+      prefill step is required again.
```
tchedaTT: Required? this conflicts with "The policy only ever chooses a decode-only step in place of a prefill step." below

README.md
```diff
@@ -312,6 +312,9 @@ Common options:
 | `fabric_reliability_mode` | Fabric reliability mode, such as `STRICT_INIT` or `RELAXED_INIT`. |
 | `dispatch_core_axis` | Dispatch core axis, `row` or `col`. |
 | `always_compat_sampling` | Use vLLM's LogitProcessor and sampler path even when not required by the batch. Default: `false`. |
+| `decode_interleave_enabled` | Insert decode-only steps into a run of consecutive prefill steps, so a running request's inter-token latency does not scale with another request's prompt length. Default: `true`. |
+| `decode_interleave_prefill_steps` | Consecutive prefill steps allowed before one insertion. Raise it to favor time to first token, lower it to favor inter-token latency. Default: `2`. |
+| `decode_interleave_decode_steps` | Decode-only steps each insertion runs before a prefill step is required again. Does not loosen the latency bound; buys more decode progress per insertion, and more cheaply than a shorter prefill run does. Default: `1`. |
```
tchedaTT: Required or allowed? Or does it only set how many decodes are inserted every time the prefill limit is reached?

tchedaTT:
With interleaving, a prefill that barely fits at admission can run out of kv cache when an advancing decode takes some. Upstream has similar problems, but maybe we should think about different scheduling behavior when kv pressure occurs.

Nitpick (codex): At 2/2, consider: two prefill steps finish the pending work, then one ordinary decode step runs with no prefill pending. A new prompt arrives. The policy still holds _prefill_run=2, _decode_run=1, so it chooses another decode immediately. This contradicts the promised behavior that a newly arriving prompt gets its first prefill step immediately. Larger decode allowances extend that delay.

viktorpusTT:
On the KV point: agreed, and it is specific to this change rather than inherited from upstream. Without interleaving a chunked prefill runs to completion before any decode advances, so the decodes cannot grow the KV footprint underneath it. With interleaving they can.

One part of it is already covered. `_schedule_decode_only` (`scheduler.py:603`) removes every `is_prefill_chunk` request from `self.running` for the duration of the inserted step (`scheduler.py:621`), so the base scheduler's preemption loop cannot see a partial prefill and cannot preempt it during an insertion.

What is not covered is the step after. The inserted decode step does allocate blocks, so the next prefill step can fail to allocate, and there the partial prefill is visible and preemptible. The existing zero-progress fallback keeps that from livelocking (a prefill pass that schedules nothing falls back to decode), but it does not stop the partial prefill losing its computed prefix to a preemption that would not have happened with the policy off.

I have not measured this, so I do not know how reachable it is in practice. Rather than guess at a guard, I would rather measure it first: one long chunked prefill plus enough concurrent decodes to drive the block pool to exhaustion, counting preemptions of the partial prefill with the policy on versus off, at a `max_model_len` and `max_num_seqs` chosen so the pool is genuinely tight. If preemptions only appear with the policy on, that quantifies the regression and tells us whether the guard needs to be "do not insert a decode step while free blocks are below some margin" or something smarter.

Happy to add that as a follow-up issue and keep it out of this PR, or to run it here before merge if you would rather know now.

tchedaTT:
Let's go with a followup issue, maybe a general one for scheduling behavior under kv pressure

docs/SCHEDULING.md
```diff
@@ -190,6 +195,137 @@ Two things follow for the TT execution path:
...
+period.
+
+Four properties keep the policy contained:
+
+- It only ever chooses decode *in place of* prefill. It never forces prefill,
+  so it cannot delay a decode that would have run anyway.
+- It only counts a request that a decode step can advance. A partial-prefill
+  continuation does not count: `_schedule_decode_only` hides it, and it samples
+  no token, so inserting a decode step on its account would produce an empty
+  step.
```
tchedaTT: Is this tryign to say that only running decodes can be inserted?

docs/SCHEDULING.md
```diff
@@ -190,6 +195,137 @@ Two things follow for the TT execution path:
...
+- The first step of any prefill run is never taken, so a prompt arriving while
+  decodes run gets its prefill step immediately. A decode step taken because
+  nothing was pending is an ordinary decode, not part of an insertion, and
+  clears both counters; counting it would leave the policy part-way through an
+  allowance and cost the next arriving prompt its first prefill step.
```
tchedaTT: Run-on sentence. Is this saying "We never replace the first step of a prefill with a decode"? Why is that important?

docs/SCHEDULING.md
```diff
@@ -190,6 +195,137 @@ Two things follow for the TT execution path:
...
+  one insertion runs. Reaching this count stops the policy choosing decode; it
+  never forces a prefill step.
+- `decode_interleave_enabled` (default `true`): the off switch, which restores
+  the strictly prefill-first policy.
+
+Both bounds are step counts, not token counts, because the interleave
+granularity is a whole step. They give a two-sided guarantee: a running decode
+advances at least once every `prefill_steps + decode_steps` steps, and pending
+prefill work gets a step at least once every `prefill_steps + decode_steps`
+steps. Neither side starves, and the policy cannot oscillate faster than that
```
tchedaTT: this dosn't seem accurate, more like "prefill can be delayed by at most decode_interleave_decode_steps"

docs/SCHEDULING.md
```diff
@@ -190,6 +195,137 @@ Two things follow for the TT execution path:
...
+spread over a longer decode run. Measured on the hardware below, the added
+time to first token per delivered decode token falls from 27.9 ms at 4 prefill
+steps with 1 decode step to 16.1 ms at 2 prefill steps with 4 decode steps.
+
+One effect is worth knowing because it is not a scheduling effect at all. Under
+async scheduling a step's sampled token is not handed to the client when the
+step finishes; it is handed over when the engine processes that step's output,
+which happens after the *next* step has been submitted. On TT a prefill step is
+a whole synchronous chunk, so an insertion's last decode token reaches the
+client a chunk later than the tokens before it.
```
tchedaTT: Is this trying to say that the earlier ones can process immediately because the followign decodes are async, and the last one needs to wait for the prefill to actually process?

docs/SCHEDULING.md
```diff
@@ -190,6 +195,137 @@ Two things follow for the TT execution path:
...
+
+| `decode_interleave_prefill_steps` | worst decode gap | median TTFT | output throughput |
+| --- | --- | --- | --- |
+| policy off | 11152 ms | 4338 ms | 138.1 tok/s |
+| 4 | 1603 ms | 4228 ms | 137.3 tok/s |
+| 2 (default) | 899 ms | 4381 ms | 136.4 tok/s |
+| 1 | 523 ms | 4660 ms | 131.5 tok/s |
+
+The same run with chunked prefill switched off entirely gives a 1853 ms worst
+decode gap, so at this token budget a run of prefill chunks makes the worst
```
tchedaTT: why is the gap smaller with chunked prefill off, than with chunked prefill on and no interleaving? wouldn't the decode have to wait for the prefill anyway in that case? is out chunked prefill that much slower than non-chunked prefill?

src/vllm_tt_plugin/platform.py
```diff
@@ -172,6 +173,28 @@ def _disable_chunked_prefill(vllm_config: "VllmConfig", reason: str) -> None:
     scheduler_config.long_prefill_token_threshold = 0
 
 
+def _report_decode_interleave_policy(vllm_config: "VllmConfig") -> None:
+    """Validate the decode-interleave settings and log the resolved policy.
+
+    Called for its raising side effect too: an invalid ``decode_interleave_*``
+    value must fail at config time, not at the first scheduler step in a worker
+    subprocess. The log line is the only external signal of the active policy.
```
tchedaTT: if it's important that this validates in addition to reporting, this function should be renamed so that it's visible at the callsite. Then, we can get rid of most of this docstring.

src/vllm_tt_plugin/scheduler.py
```diff
@@ -55,6 +59,100 @@ def from_prefill_intent(cls, prefill_intent: int) -> "TTSchedulingMode":
...
+      decode-only step is inserted.
+    - ``decode_steps``: how many decode-only steps one insertion runs before
+      the policy stops choosing decode and a prefill step may run again.
+
+    Both bounds are step counts, not token counts, because the interleave
+    granularity is a whole step. They give a two-sided guarantee: a running
+    decode advances at least once every ``prefill_steps + decode_steps`` steps,
+    and pending prefill work gets a step at least once every
+    ``prefill_steps + decode_steps`` steps, so neither side starves and the
+    policy cannot oscillate faster than that period.
```
tchedaTT: this duplicates the lines in scheduling.md, maybe we should refer to there? If not, then the same comments i added there apply here

tchedaTT (approved):
Please respond/resolve the new comments before merging

viktorpusTT:
Thanks, all addressed in 786f0ea. Follow-up issue for KV pressure is #133, filed as a general one for scheduling behaviour under KV pressure rather than an interleave-specific one; it records the partial mitigation that already exists, the cases that are not covered, and the experiment to run first.

On the bounds: you were right, and my text was loose in both directions, not just on the prefill side. The longest run without a decode step is `decode_interleave_prefill_steps` and the longest run without a prefill step is `decode_interleave_decode_steps`, so each side is bounded by the *other* count, never by their sum. Measured off the policy itself:

```
  K  D   cadence            longest P run   longest D run
  2  1   PPDPPDPPDPPDPPDP         2              1
  4  1   PPPPDPPPPDPPPPDP         4              1
  2  2   PPDDPPDDPPDDPPDD         2              2
  2  4   PPDDDDPPDDDDPPDD         2              4
```

Both copies of the wrong sentence are fixed, and the docstring no longer carries a second copy of the rationale at all.

On why the gap is smaller with chunked prefill off than on without interleaving: two effects compound and the bigger one is not chunking overhead. With chunked prefill off, `_disable_chunked_prefill` raises `max_num_batched_tokens` to `max_model_len`, so a step prefills a whole prompt, or several that fit the budget, and spaced arrivals leave gaps where pending prefill work drains and decode steps run. A decode request then waits for one step at a time. Split into chunks, the next prompt arrives before the previous one finishes, the runs merge, and the decode waits for every queued prefill instead of one. Chunking overhead is the second, smaller term; one contributor there is `get_padded_prefill_len` rounding a chunk up to 128, 1024, or the next power of two. That explanation is now in the doc, without figures.

Numerical results are out of `SCHEDULING.md` entirely, and the text points to this PR for them.

`_report_decode_interleave_policy` is now `_validate_and_log_decode_interleave_policy`, so the raising is visible at the call site and the docstring is one line.

Host tests: 103 passed. The scheduling logic is unchanged by this round, so the device results already recorded on this PR still apply; I have not re-run them.

tchedaTT:
> On why the gap is smaller with chunked prefill off than on without interleaving: two effects compound and the bigger one is not chunking overhead. With chunked prefill off, `_disable_chunked_prefill` raises `max_num_batched_tokens` to `max_model_len`, so a step prefills a whole prompt, or several that fit the budget, and spaced arrivals leave gaps where pending prefill work drains and decode steps run. A decode request then waits for one step at a time. Split into chunks, the next prompt arrives before the previous one finishes, the runs merge, and the decode waits for every queued prefill instead of one. Chunking overhead is the second, smaller term; one contributor there is `get_padded_prefill_len` rounding a chunk up to 128, 1024, or the next power of two. That explanation is now in the doc, without figures.

I'm not convinced by this explanation. What scenario causes consecutive chunked prefills and wouldn't also cause consecutive prefills for non-chunked?

viktorpusTT:
> > On why the gap is smaller with chunked prefill off than on without interleaving: two effects compound and the bigger one is not chunking overhead. With chunked prefill off, `_disable_chunked_prefill` raises `max_num_batched_tokens` to `max_model_len`, so a step prefills a whole prompt, or several that fit the budget, and spaced arrivals leave gaps where pending prefill work drains and decode steps run. A decode request then waits for one step at a time. Split into chunks, the next prompt arrives before the previous one finishes, the runs merge, and the decode waits for every queued prefill instead of one. Chunking overhead is the second, smaller term; one contributor there is `get_padded_prefill_len` rounding a chunk up to 128, 1024, or the next power of two. That explanation is now in the doc, without figures.
> 
> I'm not convinced by this explanation. What scenario causes consecutive chunked prefills and wouldn't also cause consecutive prefills for non-chunked?

The only thing chunking changes in that case is how long one prefill takes. In E2 that is the whole story. One 16384-token prompt takes 1870.8 ms unchunked and 2889.3 ms at budget 2048, against prefillers arriving 2000 ms apart. Under the spacing, each prefill finishes before the next prompt arrives and decodes run in between, so each decoder sees 4 gaps of about one prefill each, worst 1853.1 ms. Over it, the queue never drains and the four prefills merge into one gap, worst 11152.5 ms against 4 x 2889.3 = 11557 ms.

tchedaTT:
That's a 50% overhead from chunking - is it always that bad?

viktorpusTT:
Two likely mechanisms here:
1. TT Transformers distributes 32 query heads across eight chips: 32 / 8 = 4 query heads per chip. Each chip processes all 2,048 token positions for its four heads.
SDPA divides each head’s query positions using q_chunk_size=256: 2048 / 256 = 8 query chunks per head.
SDPA groups those query chunks into pairs: 8 / 2 = 4 pairs per head.
With four heads, that gives 4 × 4 = 16 pairs per chip. SDPA assigns each pair to one core: The compute grid contains 64 cores within each chip, but only 16 cores receive query work.
Compared to 8,192 tokens: 64 pairs, assigned to all 64 cores. 16,384 tokens: 128 pairs, assigned to all 64 cores, two pairs per core.
(All of this for Llama-3.1-8B on T3K.)

2. Intermediate chunks compute output that vLLM discards. Each intermediate scheduler chunk still performs final normalization and LMHead, converts vocabulary logits, copies those logits to CPU, and synchronizes. TTModelRunner suppresses the intermediate output afterward. Eight chunks therefore add seven unnecessary vocabulary projections and readbacks.

Consistent with the measured:
* No chunking: 1,870.8 ms to first token.
* Two 8,192-token chunks: 1,899.1 ms, or +1.5%.
* Eight 2,048-token chunks: 2,889.3 ms, or +54.4%.

Conclusions: 
* Chunking without interleaving saves memory for sliding window attention.
* Chunking with interleaving also allows tuning TTFT vs TPOT gaps. Compromises, no net positive.
* Results are model- and device- dependent.

PR #120, opened 2026-09-10
[feat] Speculative decoding: contract types and a conformant host stand-in model

viktorpusTT:
## TL;DR

Adds the speculative-decoding contract types (`SpecPlan`, `SpecReject`, `DraftOutput`, `VerifyOutput`) and a host stand-in model class that implements the contract in torch. Nothing in the plugin consumes them yet: this is the surface the runner-side work is built against, landing ahead of its consumer on purpose.

## Details

The plugin rejects every speculative request today (`platform.py`, `_apply_check_and_update_config`). The runner-side work needed to lift that is almost entirely host logic: config admission, `[B, 1+K]` input construction, scheduler bookkeeping, the accept walk, output emission. None of it needs a device to exercise, but all of it needs a model that implements the contract, and no tt-metal model does yet.

So this PR adds that model as a test double, and the contract types it returns.

`src/vllm_tt_plugin/spec_decode.py` holds the wire surface and the validation of the values that cross it. It reads no `model_capabilities` key and admits no configuration, which keeps it inside the section 6 rule in `AGENTS.md`: a plugin change that consumes a new capability needs the matching tt-metal declaration in the same pair of PRs. Nothing here consumes one. `normalize_declared_values` validates a declaration a caller has already read, and lives here next to the constant sets it validates.

Two contract properties are enforced by the types rather than left to each caller, because leaving either to a caller invites an error that only surfaces inside an accept walk:

- `SpecPlan.accepted_counts_range` is the inclusive `[1, 1+K]` domain of `accepted_counts`. It is a count of committed tokens and never zero, so a model selecting a per-candidate state slot selects slot `accepted_counts - 1`. This matches upstream's `num_accepted_tokens` convention, which fills with 1 for new and unused rows (`vllm/v1/worker/gpu_model_runner.py`, `vllm/v1/worker/gpu_input_batch.py`) and reads slot `n_acc - 1` in its recurrent kernels (`vllm/model_executor/layers/fla/ops/fused_recurrent.py`).
- `VerifyOutput.__post_init__` enforces the pairing between a `spec_mode` and the fields its return must carry. `MODE_REQUIRED_FIELDS` is public because it is contract information, so a caller can check a return it did not build.
- `SpecPlan` refuses `drafter_target_cache_requires` on a `drafter_state` that reserves bytes. The two cost-bearing states describe what the runner must budget; `"shared_with_target"` describes a drafter that reserves nothing and instead constrains how the target's caches are organised. A requirement on a cost-bearing state is a contradiction, not extra information, and the failure mode it guards against is reading the wrong cache positions rather than failing to allocate.
- `PLACEHOLDER_TOKEN_ID` names the tail marker for a fixed-width block that committed fewer tokens than its width. A real token identifier cannot serve as the pad, including the end-of-sequence identifier, because it is indistinguishable from a committed token and silently turns a short commit into a stop.

`tests/spec/fake_spec_model.py` is the stand-in. Its value is that it refuses every input the contract forbids rather than tolerating it: a ragged committed block, a mis-ranked tensor, an empty batch, a width other than `1+K`, a non-int32 or out-of-domain `accepted_counts`, a per-row draft count outside `[0, K]`, and a `None` count outside the one case that permits it. That makes the follow-up runner work fail here rather than in a debugger.

Acceptance in the stand-in is per row, capped at `min(accept_depth, num_valid_drafts[i])`, and never reduced across the batch. A batch-wide cap would let one grammar-truncated request destroy every other request's speculation, which the contract forbids, and a stand-in that did so could not drive the mixed-row case the runner has to get right.

Both `argmax_ids` and `logits` are served, not just the mode a first model would land, because the mode dispatch is the runner code most likely to be wrong and a second mode costs nothing in a stand-in. A test asserts the two return the same ids, so a test of one is evidence about the other. `fused_sample` is refused, because no implementation serves it today on either side; the contract keeps the mode for a model that will, and a stand-in that served it would give a passing test for a path nothing can run. `SpecPlan` still accepts the mode, because refusing a mode no model implements is a config-time policy and not a property of a value object.

`spec_plan` is a classmethod, since the runner calls it before an instance exists. `make_fake_spec_model(**knobs)` returns a configured subclass rather than mutating class attributes, so one test cannot configure the next, and it refuses a knob naming a contract method rather than replacing a primitive with a plain value. `decode_forward` takes `spec_mode` with no default, so a caller that forgets it fails instead of silently receiving greedy ids.

## Related Artifacts

Contract: tenstorrent/vllm-tt-plugin#110. Model-side feature request: tenstorrent/vllm#442. First tt-metal implementation: tenstorrent/tt-metal#55548.

This is step 1 of a plugin-side plan whose later steps consume these types: config admission, the `[B, 1+K]` input build, an ngram drafter with a greedy accept walk, a torch accept walk for sampled requests, and structured output over drafts.

No paired tt-metal change is needed, because this PR consumes no capability.

## Validation

Host-only, no Tenstorrent hardware, using the `AGENTS.md` recipe.

```text
$ PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/spec -q
71 passed in 3.12s

$ PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/ --ignore=tests/tt -q
456 passed, 14 warnings in 3.65s

baseline on origin/main, excluding tests/spec:
$ PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/ --ignore=tests/tt --ignore=tests/spec -q
385 passed, 14 warnings in 3.64s
```

`pre-commit run` over the four files:

```text
check toml...........................................(no files to check)Skipped
```

The conformance tests were mutation-checked, so they are known to bite rather than to pass trivially. Each mutation breaks one property the suite claims to pin:

```text
per-row cap -> batch-wide minimum:            2 failed, 61 passed
block shape-equality guard removed:           1 failed, 62 passed
block rank guard removed:                     1 failed, 62 passed
accepted_counts upper bound made exclusive:   1 failed, 62 passed
empty-batch guard removed:                    2 failed, 61 passed
a mode dropped from MODE_REQUIRED_FIELDS:     3 failed, 60 passed
all mutations reverted:                      63 passed
```

The third commit tracks three contract changes and adds three more mutations:

```text
cost-vs-constraint contradiction check dropped:   1 failed, 70 passed
shared_with_target dropped from DRAFTER_STATES:   2 failed, 69 passed
PLACEHOLDER_TOKEN_ID set to a real token id:      1 failed, 70 passed
all mutations reverted:                          71 passed
```

The first four of those mutations survived the suite as originally pushed, and the last two properties were unpinned entirely. The second commit fixes the per-row defect they exposed and closes all six gaps, so the mutation table above is also the review record.

No device validation applies: this PR adds no device-reaching code, and `ci/host-stubs/ttnn/` would fail loudly if it did.

## Checklist

### DIFF
--- a/src/vllm_tt_plugin/spec_decode.py
+++ b/src/vllm_tt_plugin/spec_decode.py
@@ -3,16 +3,22 @@
 """Contract types for speculative decoding between the runner and a TT model.
 
 The contract these types encode is specified in
-https://github.com/tenstorrent/vllm-tt-plugin/issues/110. Nothing here reads a
-`model_capabilities` key or admits a configuration; this module is the wire
-surface only, so a model class and the runner can agree on shapes and modes
-before either side implements a step of the loop.
+https://github.com/tenstorrent/vllm-tt-plugin/issues/110. This module holds the
+wire surface and the validation of the values that cross it, so a model class
+and the runner can agree on shapes and modes before either side implements a
+step of the loop. It reads no ``model_capabilities`` key and admits no
+configuration: ``normalize_declared_values`` validates a declaration a caller
+has already read, and lives here next to the constant sets it validates.
 
-Two properties of the contract are enforced here rather than left to each
-caller, because both were ambiguous enough in earlier revisions to produce
-off-by-one and shape errors: the inclusive range a valid ``accepted_counts``
-entry lies in, and the pairing between a ``spec_mode`` and the fields a verify
-return must carry for it.
+Per the contract, one step is verify then propose: the runner calls
+``decode_forward`` over the ``[B, 1+K]`` candidate block, walks acceptance, and
+calls ``propose_draft_tokens`` with that same step's hidden state.
+
+Two properties are enforced by these types rather than left to each caller,
+because leaving either to a caller invites an off-by-one or a shape error that
+only surfaces inside an accept walk: the inclusive range a valid
+``accepted_counts`` entry lies in, and the pairing between a ``spec_mode`` and
+the fields a verify return must carry for it.
 """
 
 from collections.abc import Sequence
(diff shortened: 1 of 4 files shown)


viktorpusTT:
Merged without review as this is not affecting any live path. Unblocking next steps.

PR #125, opened 2026-09-14
[feat] Speculative decoding: admit or refuse a configuration at config time

viktorpusTT:
## TL;DR

Resolves a speculative configuration against the model's own `spec_plan` at configuration time, publishes the resolved plan, and then refuses the launch because this branch alone has no execution path for it. #130 removes that refusal along with the function behind it. Every refusal raises with the offending values and the flag that changes them.

**Do not merge yet**, and the reason has narrowed to one gate: no tt-metal
model declares the capabilities this reads, so `AGENTS.md` section 6's pairing
rule applies. The other gate is answered on top of this branch: #127 builds the
candidate block and #130 drives the loop, which is what deletes the refusal
this branch adds. The speculative test model in tt-metal's
`models/vllm_test_utils/` is what satisfies the remaining gate without touching
a production model.

## Details

`TTPlatform._apply_check_and_update_config` previously refused every speculative launch with a blanket assert. It now runs a resolution pass after the lane fold, where the final concurrency and the stored lane count are both available.

`src/vllm_tt_plugin/spec_admission.py` holds the policy. `resolve_speculative_plan` returns the model's `SpecPlan` when a configuration is serviceable and `None` when none was requested, and raises otherwise. It refuses when the model does not declare `supports_spec_decode`; when the speculative method is one the backend cannot serve; when the model's `spec_requirements` lacks what the method needs; when a hidden-feed method finds no `spec_hidden_handoff`; when the method needs a scheduler-owned drafter cache; when `spec_plan` is absent or not callable; when it returns `SpecReject`, whose `reason` and `supported_k` are quoted; when it returns a plan above the requested draft length, a paged drafter state, or no accept mode the runner can drive.

`src/vllm_tt_plugin/spec_decode.py` stays a leaf contract module: it reads no capability key, admits nothing, and imports nothing from vLLM at run time, so a tt-metal model class can import the value types without pulling in plugin policy.

**The plugin owns the method-to-requirement table.** A model declares requirements such as `device_propose` and `hidden_feed` and never a vLLM method name, so a new upstream method is a plugin change rather than a model release. The table is built from vLLM's own `EagleModelTypes` literal, so an upstream rename drops a name out of the table rather than leaving the plugin mapping a name vLLM no longer knows.

**The resolved plan is not stored as `output_tokens_per_step`.** `is_tt_block_output_model` is `output_tokens_per_step > 1`, and that predicate selects the block-output rail, which neutralizes HTTP sampling controls, disables logprobs, rejects resumable requests and switches the scheduler to block placeholder accounting. Speculation honours sampling through its `logits` mode. The plan goes under its own key, as a dictionary, because `VllmConfig.compute_hash` calls `json.dumps` over `additional_config`. The same collision is why a model declaring a block width cannot also speculate; that conflict is raised in `platform.py` beside the other contradictions involving that rail.

**A reduced draft length is published back.** When a model's `spec_plan` serves less than was requested, `speculative_config.num_speculative_tokens` is rewritten and the reduction logged, because vLLM's own `Scheduler` derives its lookahead-slot reservation from that field.

**An otherwise valid launch is refused.** vLLM's `EngineCore` calls `take_draft_token_ids` on the executor on every step of a speculative run, and `TTWorker` implements none; `TTModelRunner` drives no verify-then-propose loop. A launch that passed resolution would therefore raise `AttributeError` on its first step, or, with a stub returning `None`, serve no speculation while reporting that it does. Resolution runs in full first so a model author sees their own declaration error, and one call refuses the launch. That call is the only thing to delete when the execution path lands.

`docs/SPEC_DECODE_CONTRACT.md` specifies what a tt-metal model must implement: the four capability entries, the `spec_plan` signature, what it may not read off `vllm_config`, and the rules admission enforces. `AGENTS.md` section 6 lists the three new keys and the new method contract.

## Related Artifacts

Contract: tenstorrent/vllm-tt-plugin#110. Step 1: tenstorrent/vllm-tt-plugin#120. Model-side feature request: tenstorrent/vllm#442.

## Validation

Host-only, no Tenstorrent hardware, using the `AGENTS.md` recipe.

```text
$ PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/spec -q
117 passed in 4.10s

$ PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/ --ignore=tests/tt -q
502 passed, 14 warnings in 4.51s

baseline on this branch without the two new test files:
457 passed, 14 warnings in 4.02s
```

`pre-commit run` on the staged files:

```text
check toml...........................................(no files to check)Skipped
```

Mutation-checked, because a silently admitted configuration is the dangerous failure here:

```text
delete the whole admission call site:            8 failed, 492 passed
drop the execution refusal:                      5 failed, 495 passed
drop the effective_k writeback:                  1 failed, 499 passed
drop the paged-requirement refusal:              1 failed, 116 passed (tests/spec)
is-None instead of callable for spec_plan:       1 failed, 499 passed
resolve before the lane fold:                    1 failed, 499 passed
all mutations reverted:                        502 passed
```

No device validation applies: this adds no device-reaching code, and `ci/host-stubs/ttnn/` would fail loudly if it did.

### There is no metal counterpart yet

It reads `supports_spec_decode`, `spec_requirements` and `spec_hidden_handoff`, which no tt-metal model declares. The paired tt-metal change is a `spec_plan` classmethod plus those three keys on a model class; it is not open yet.
Every launch it resolves is refused. Merging now adds a configuration-time refusal and no capability.

It is open for review now because it is fully testable against the stand-in model merged in #120, and because reviewing the contract surface before a model implements it is the cheapest time to find an error in it.

## Checklist


docs/SPEC_DECODE_CONTRACT.md
```diff
@@ -0,0 +1,93 @@
...
+https://github.com/tenstorrent/vllm-tt-plugin/issues/110.
+
+## Status
+
+The plugin admits a configuration and then refuses the launch, because no
+execution path exists yet: `TTWorker` implements no `take_draft_token_ids` and
+`TTModelRunner` drives no verify-then-propose loop. Implementing this contract
+on a model is therefore useful now for validating declarations, and does not
+yet produce a speculating server.
+
```
viktorpusTT:
```suggestion
```

PR #131, opened 2026-09-16
[feat] Speculative decoding: the model's own drafter proposes on device

viktorpusTT:
## TL;DR

The model's own drafter can now propose on device. `TTModelRunner` calls
`propose_draft_tokens` after each commit and hands it that step's verify hidden
handle, so a model with an MTP or EAGLE head drafts for itself instead of
relying on the host n-gram proposer. Neither first production model drafts with
n-gram, so this is the path they need.

A speculative step also runs under asynchronous scheduling now: acceptance is
walked where the readback completes, and the commit and the next proposal run
on the engine thread at the next step's drain. Admission requires a model to
declare it. See "Speculative steps under asynchronous scheduling" below.

Paired with `tt-metal#56646`, which implements the drafter in
`DummySpecDecodeModel` and declares the asynchronous capability. Based on
`viktorpus/spec-decode-ngram-greedy` (#130).

## Why `custom_class` is the vLLM-side method name

vLLM validates `method` against a `Literal`, so the plugin cannot invent a
name, and every device-drafter name upstream offers wants something a TT model
cannot give:

| method | what vLLM demands |
| --- | --- |
| `eagle`, `eagle3`, `medusa`, `mlp_speculator` | a `model` path to load as the draft head |
| `mtp` and every `*_mtp` | a draft architecture resolved from the target's `model_type` |
| `draft_model` | a scheduler-owned paged drafter cache, which TT does not allocate |
| `extract_hidden_states` | a hidden-state dumping mode, and it changes the computation graph |
| `custom_class` | a dotted proposer path, and nothing else |

`custom_class` is also the truthful one: it means a proposer vLLM does not own,
which is exactly a model-owned drafter. vLLM sets `draft_model_config` to the
target's and loads nothing. Its `model` key is pinned to
`vllm_tt_plugin.model_owned_drafter`, a path nothing imports, because any other
value would name a proposer that never runs and read as the thing drafting.

## The propose call

```python
def propose_draft_tokens(
    self,
    num_drafts,            # K
    committed_tokens,      # [B, 1+K] int32, this step's committed block
    committed_positions,   # [B, 1+K] int32
    accepted_counts,       # [B] int32 in [1, 1+K]
    hidden=None,           # the HiddenHandle this step's verify returned
) -> DraftOutput
```

Rows are the verify's rows, padding included, because a drafter's state is
indexed by row and a device graph has one shape. That is why the accept walk
now runs over the full padded block and only the live rows reach a request.
Which entry of the committed block is a row's last token is
`accepted_counts - 1`, the same arithmetic the verify uses to select a candidate
state slot.

`hidden` travels on `TTDecodeSubmission` rather than in runner state, so it
cannot outlive the step it belongs to. The runner never interprets it.

Refusals, all by type or by name: a drafter returning something other than a
`DraftOutput`, a wrong-shaped draft tensor, a tensor whose dtype is not int32,
and a draft id outside the vocabulary. The last two matter because a stored
draft is verified next step and committed if the model agrees with it, and
because the ids are read out with `int()`, which would truncate a float
silently.

Two shapes the runner normalizes so a model sees one contract. A padding row's
committed positions stay negative rather than becoming `0, 1, 2`, because the
drafter is handed the verify's rows with no live-row mask and the position is
the only marker for a row no request owns. And a narrow verify's one-column
committed block is padded to `1+K` before the propose call, so a model that
declares `supports_narrow_decode` still sees the documented width here.

## Speculative steps under asynchronous scheduling

Asynchronous scheduling splits a step: `execute_model` submits and returns
nothing, the engine collects the output later on a thread that is not the
engine thread, and the runner applies it to request state at the top of the
following step. Acceptance did not fit that split, because `_finish_spec_decode`
walked, committed and published in one call on the engine thread. It is now
split at the same seam:

| method | where it runs | what it touches |
| --- | --- | --- |
| `walk_spec_acceptance` | where the readback completes | nothing host-side: validation and the accept walk |
| `spec_committed_prefixes` | where the readback completes | the step's own positions, to clip each prefix at `max_model_len` |
| `build_spec_runner_output` | where the readback completes | builds the `ModelRunnerOutput` the engine reads |
| `commit_spec_acceptance` | the engine thread, at the next step's drain | request histories, accepted counts, the next proposal |

The synchronous tail calls the four in a row, so there is one accept walk in
the plugin and not a second copy for the deferred path.

Ordering, which is the part that is not a throughput claim: a speculative step
registers with `overlap_ok=False`, so `must_drain_pending_async_steps` is true
while it is outstanding, `execute_model` calls
`wait_for_all_pending_async_steps` before `build_model_input`, and
`build_model_input` applies the drained result before `_prepare_model_inputs`
builds anything. A verify is therefore never submitted while the previous
verify's acceptance is unapplied, even under vLLM's `step_with_batch_queue`,
which schedules the next batch before resolving the prior future. Verify,
accept and propose stay serialized per request. What asynchronous scheduling
buys here is the engine's own overlap, not two verifies in flight.

Lifecycle: a request that finished, was cancelled, was preempted, had its row
taken by another request, or had its frame marked stale by a forced
prefix-cache reset between a step's submission and its commit gets the
treatment that case calls for, and three defects in that came out of testing
it:

- **A late commit revived a preempted request's candidate state.** An ordinary
  preemption keeps the tokens (the engine has them, and the resume restores
  from that history) but releases the request's row and its
  `_req_state_slot` entry. The commit wrote `_req_accepted_counts` anyway and
  drafted a continuation, so a resumed request would have reached the model
  claiming a candidate state slot the model no longer held. Now a request with
  no row at commit time keeps its tokens and records neither.
- **The model drafter sized each row's remaining context by the step's row.**
  Correct for the committed block and the counts, which the model answered in
  step-row order; wrong for the persistent batch, because a completion between
  submission and commit makes `condense` slide a request into another row and
  the row it left keeps a stale length. Now resolved through
  `req_id_to_index`, the way the commit already did.
- **A frame a forced prefix-cache reset marked stale was not applied, as
  intended, but was still drafted from**, leaving a draft that continued tokens
  the scheduler had discarded. The skip set now reaches the proposal.

All three are reachable only when a commit runs a step after its submission,
which is why the synchronous path never showed them.

Admission: a model must declare
`model_capabilities['supports_async_spec_decode']`. Two demands make that a
separate declaration rather than the conjunction of `supports_async_decode`
and `supports_spec_decode`:

1. `read_decode_output` is handed a verify. On a synchronous speculative step
   the hook is never called, because the runner asks for the output with the
   submission. On this path it is called with the tensor unwrapped from the
   `VerifyOutput`, so it receives the mode's `[B, 1+K]` block and the number of
   tokens per row committed out of it is decided by the host after the forward
   returns.
2. The verify's hidden handle is held across the readback and a queue, and
   reaches `propose_draft_tokens` at the next step's drain. A model declaring
   `spec_hidden_handoff: ["on_device"]` is promising the device state behind it
   is still valid then.

The decode reload contract that `supports_async_decode` answers to covers
neither: its requirements are written for a decode that commits one token per
forward, speaking of a persistent token buffer holding the selected token and
one resident position advance per forward. A model can satisfy all of them for
an ordinary decode and be wrong for a deferred verify, and deriving the pairing
would enlarge what the models already declaring it promised.

Absent the declaration, the launch is refused by name and the message names
`--no-async-scheduling`. `supports_async_decode` is still read first, so a
model declaring only the speculative half keeps decoding synchronously.

One upstream blocker worth recording: vLLM 0.26.0 refuses asynchronous
scheduling with `custom_class` before the plugin sees the configuration.
`VllmConfig.__post_init__` admits asynchronous scheduling only for EAGLE, MTP,
a draft model, NGram GPU and DSpark, and it runs before
`current_platform.check_and_update_config`. So on a real server the pairing is
reachable only under an MTP-family method name. The host tests drive it through
the runner's real entry points instead.

## Ordinary batched decoding inside a speculating server

The objection this answers: an engine running several requests against a
baseline model loses throughput without asynchronous scheduling. It did, and
worse than reported. `_prepare_model_inputs` set `spec_mode` on **every**
decode step of a speculating launch, and a step carrying `spec_mode` is
registered as not overlap-safe, so a server with speculation configured never
overlapped a single decode step whatever its drafter did. A request drafting
nothing still paid for a `[B, 1+K]` forward per step to verify nothing.

Three changes, and the first two are contract changes:

**`DraftOutput.num_valid`**, `[B]` int32 and optional, is how a drafter
declines: a row at 0 is drafted for nowhere. There was no way to say that, so
an adaptive drafter had to fabricate tokens it did not want verified. `None`
means all `K`, so every existing drafter keeps working. A declining row's ids
are no longer range-checked either, since nothing reads them.

**A step with nothing to verify is an ordinary decode.** A step is a verify
when some row carries a draft, or when some row's previous step committed more
than one token: `accepted_counts` is how a model finds which candidate state
slot that commit landed on, so the step after it carries the count even when it
drafts nothing, and one step resolves it. Otherwise, for a model declaring
`supports_narrow_decode`, the step is sent as that model's own decode call, with
no `spec_mode` and neither side tensor, and it is overlap-safe under the
conditions that govern every other decode. Leaving speculation costs exactly
one verify per transition. The documented narrow call, a one-column
`VerifyOutput`, is gone: nothing implemented it (every
`supports_narrow_decode` in either repository was `False`), and it could not
overlap either, being a verify. A declaring model now implements two calls and
no third shape.

**The proposer is still asked on those steps**, or a launch that went draftless
once would never draft again. The committed block is one column wide and padded
to the uniform `1+K` before the call. On an asynchronous launch this runs at
the next step's drain, because the drafts continue a token that had to be read
back first, so a drafter that starts offering again is acted on one step later.
That closes this path to a drafter fed its hidden state through the runner,
since such a step returns no handle: `load_model` keeps that pairing on the
verify path and logs why. Which in turn exposed an over-demand, now fixed: the
`custom_class` method required `hidden_feed` of every model, though what its
drafter reads is the model's own statement.

One related defect: declining for a row left the previous step's offer standing
in `_proposed_draft_token_ids`, because both proposers only wrote entries.
Unreachable in production, where `take_draft_token_ids` drains the map every
step, and now erased explicitly.

Paired with `tt-metal#56646`'s `TT_SPEC_DRAFT_POLICY=solo`, which offers the
full draft length while one request is live and nothing while more are, and
counts live requests by their committed positions rather than by rows.

## The asynchronous launch, reachable

Everything above was driven through the runner's entry points, because a
speculating launch could not keep asynchronous scheduling. Two things stood in
the way.

**vLLM decides before the plugin is asked.** `VllmConfig.__post_init__` raises
for an explicit `--async-scheduling` with any method outside
EAGLE/MTP/draft_model/NGram GPU/DSpark, and the automatic selection rewrites
the setting to `False` for the same set, both before
`current_platform.check_and_update_config`. `custom_class` is the only name
upstream accepts for a proposer it does not own. Both predicates read one
module-level name, `EagleModelTypes`, which `vllm.config.vllm` uses in exactly
those two places, so `_install_tt_async_spec_method_patch` rebinds that name
and changes nothing else anywhere: every other consumer imports the type into
its own namespace. Installed from `pre_register_and_update`, which is
`create_engine_config`'s first statement. Installation first checks that the
gate still makes that comparison twice and refuses the launch otherwise,
because a rebind that stops mattering would quietly serve synchronously.

**With async on, drafts stop arriving through the scheduler.**
`EngineCore.post_step` does not call `take_draft_token_ids` at all when
asynchronous scheduling is on; upstream substitutes its own ids in the worker.
`TTScheduler` was clearing `AsyncScheduler`'s `[-1]` placeholders, which on
that path are the only lookahead budget a request gets. So the admitted launch
would have decoded every step plainly and never speculated. `TTScheduler` now
keeps them under async, and `_drafts_to_verify` reads them as the reservation
they are and verifies the proposal the runner holds, trimmed to it and
consumed once. The synchronous path is untouched.

`tests/spec/test_spec_async_admission.py` is the acceptance suite, and the
first here to build real `ModelConfig`, `SchedulerConfig`, `SpeculativeConfig`
and `VllmConfig` objects rather than a `SimpleNamespace`, because the decision
under test is upstream's. Fourteen tests: the unpatched refusal in both forms,
the patched launch keeping async explicitly and automatically,
`--no-async-scheduling` still disabling, the method and sentinel surviving, a
foreign proposer path and `ngram` still refused, the padded-drafter refusal
still applying, idempotent installation, and a restructured gate failing the
launch.

Also fixed on the way: `plan_decode_reload` would tell a model to decode a
**verify** from resident inputs on a steady device-sampling step, though a
candidate block is assembled fresh from that step's drafts.

Paired with `tt-metal#56646`, where the dummy now implements the ordinary
asynchronous decode it needs to be scheduled that way: the four update
commands honored independently, resident tokens, positions and page tables,
token feedback, one position advance per forward, and a readback split from
submission.

## Validation

AI assistance (Claude Code) was used for this change.

### On a device

`wh-lb-77`, one Wormhole chip, `MESH_DEVICE="(1, 1)"`, vLLM 0.26.0, K=5, prompt
1024 tokens, 512 completion tokens per request. Every point HTTP 200 with no
invalid token id. Full report in the plan's companion,
`spec-decode-model-drafter-run.md`.

Acceptance is exactly what the model's accept depth configures, read from
vLLM's own counters, and **every step speculates** where n-gram stalled:

| accept depth | speculative steps | drafts offered | accepted | per-position | ms/step | tok/step |
| --- | --- | --- | --- | --- | --- | --- |
| 0 | 510 | 2550 | 0 | 0,0,0,0,0 | 1.286 | 1.002 |
| 2 | 170 | 850 | 340 | 170,170,0,0,0 | 1.269 | 2.994 |
| accept all | 85 | 425 | 425 | 85,85,85,85,85 | 1.307 | 5.953 |

The per-step host cost is flat in the committed width, so the per-token cost
falls with it: 1.283, 0.424 and 0.219 ms per committed token. Against the same
model's plain decode step at 1.283 ms per token, accepting five drafts is 5.8
times cheaper per token in host time.

Concurrency, accepting every draft, one server per point:

| `max_num_seqs` | requests | ms per step | tokens per step | host tokens/s | wall |
| --- | --- | --- | --- | --- | --- |
| 1 | 1 | 1.069 | 5.95 | 5,569 | 0.114s |
| 4 | 4 | 1.396 | 23.81 | 17,063 | 0.146s |
| 8 | 8 | 1.579 | 47.63 | 30,162 | 0.184s |
| 32 | 32 | 3.407 | 190.51 | 55,917 | 0.383s |

Every request took exactly 85 speculative steps at every concurrency, so no row
was starved. The fixed per-step cost amortizes across rows: 32 rows cost 3.2
times one row per step and deliver 32 times the tokens.

### The asynchronous launch, on a device

`wh-lb-77`, plugin `59c07d7`, tt-metal `1fa6deb32a0` with the `spec_test`
overlay, vLLM 0.26.0, K=5, the fixed target, the adaptive draft policy,
`sample_on_device_mode=decode_only`, and no `--no-async-scheduling`:

```text
7 passed in 6.59s
Asynchronous scheduling is enabled.
TT async decode: 128 submission(s) overlapped an outstanding step,
                 0 of them a step that was not overlap-safe
```

| Claim | What says so |
| --- | --- |
| the resolved server schedules asynchronously | vLLM's own log line, the outcome rather than the flag it was asked for |
| ordinary steps overlap | 128 submissions found a step already outstanding; the counter logs at every power of two |
| verify steps serialize | 0 of those were not overlap-safe, for the life of the server |
| speculation happens and resumes | a lone request: 55 accepted, per-position 11 at all five positions, mean acceptance length 5.58, and acceptance moved again after a peer came and went |
| a batch is not drafted for | four rows over 65 steps: 0 accepted, every per-position count 0 |
| the output is the reference | the whole completion equals the fixed target's closed form, token for token |

The first run of this configuration failed three tests, which is what it is
for. The launch could not overlap at all without `sample_on_device_mode`,
because the steady-decode fast path refuses a host-sampled step. One real
defect surfaced that no host test had: the drain-before-a-verify decision took
the placeholder list for drafts, and that list is on every scheduled request of
every asynchronous step, so every step drained and the counter stayed at zero.
And two assertions of mine were wrong: they read
`spec_decode_num_drafts_total`, which under async counts the reservation rather
than real drafts, and they waited a tenth of a second before posting a peer,
which is a hundred steps for a host-arithmetic model.

### Host overhead, measured

`tests/tt/spec/bench_spec_overhead.py` launches one server per configuration, discards a warmup interval, and measures three intervals of 8 concurrent requests at 1900 tokens each. It reports committed tokens per second from `vllm:generation_tokens_total`, host CPU from `utime` plus `stime` over the server's whole process tree in `/proc`, and the runner's own submission counters. `wh-lb-77`, plugin `5744013`, tt-metal `4b34a8f645a`, K=5, the fixed target, `sample_on_device_mode=decode_only`.

**A batch that never speculates pays 0.73 ms of host time per decode step for the drafter being configured.** An `ordinary-async` server with no `speculative_config` and an `adaptive-async` server whose drafter declines for every row of an 8-row batch did provably identical work: both committed 15200 tokens per interval, both submitted 1920 ordinary decodes and 0 verifies, and both overlapped 1916 of those 1920. Throughput was 8504.9 against 4782.9 committed tokens/s, which is 0.941 ms against 1.673 ms of wall clock per step. Host CPU accounts for 0.705 ms of the 0.732 ms difference, so the cost is work being done and not time being waited: the overlap counts are identical, so no overlap was lost. Two runs thirty minutes apart give 43.8% and 42.6%.

**Asynchronous scheduling costs throughput on an all-verify workload, as designed.** `contract-async-*` is 6.9% slower than `contract-sync-*` at accept depth 0 and 3.8% slower at accept depth 2, against a 1.0% to 1.4% spread across repetitions. At full acceptance the measured 4.9% is inside that configuration's own 7.8% spread and is not resolved by this run. None of the three `contract-async-*` server logs contains the overlap line at all, and that line prints at every power of two including 1, so `_overlapped_submissions` stayed at 0 for the life of all three: a verify is never overlap-safe, so asynchronous scheduling had nothing to overlap and charged its bookkeeping for nothing.

**What acceptance buys, in host cost.** Host CPU per committed token falls from 0.162 ms for ordinary decoding to 0.254 ms at accept depth 0, 0.096 ms at accept depth 2 and 0.056 ms at full acceptance. Throughput at full acceptance is 2.97 times ordinary decoding, against a ceiling of 6, because a verify step that commits 6 tokens per row costs 2.02 times an ordinary decode step. A drafter whose drafts are always rejected costs 44% of aggregate throughput and returns nothing.

Startup was 31.2 s to 32.3 s across all ten launches with no separation between speculating and non-speculating or synchronous and asynchronous, so configuring speculation adds no measurable startup cost.

Two counters were added for this, because vLLM's speculative metrics cannot answer it: `AsyncScheduler` reserves `[-1] * num_speculative_tokens` for every scheduled request, and `spec_decode_num_drafts_total` counts that reservation, so it moves on a server that verifies nothing. `count_decode_submission` counts real ordinary decode and verify submissions in `submit_decode`, the one funnel both execution modes share, and repeats the overlap total on the same fixed cadence because `_log_overlap_counters` reports only at powers of two and so cannot resolve an overlap count inside a measured interval.

**Scope.** The `spec_test` target computes its output on the host and performs no device compute, so every number here is a host cost and the per-step deltas are the portable figures. The percentages are relative to a step that costs about a millisecond, and a real TT decode step does not. The 2.97 factor is what the host permits, not a projected speculative speedup: a real target's verify evaluates 1 plus K positions per row on the device. No artificial completion delay is used in any configuration.

### A device regression suite

`tests/tt/spec/` turns the device experiment into reusable assertions, with
`run_spec_regression.sh` as the driver and `README.md` as the recipe. It needs
a reserved device (the worker opens a mesh whatever the model does) and
`tests/tt` is excluded from continuous integration, so this is a manual job
today.

Five configurations, measured on `wh-lb-77`, and re-run there at `9ac3451`
after the accept path was split, because the split changed code the
synchronous device path runs:

| configuration | result | after the accept-path split | after the lifecycle fixes |
| --- | --- | --- | --- |
| accept-all | 16 passed | 16 passed | 16 passed |
| accept-2 | 16 passed | 16 passed | 16 passed |
| accept-0 | 11 passed, 5 skipped at a committed width of one | 11 passed, 5 skipped | 11 passed, 5 skipped |
| capacity | 3 passed, across 6 preemptions | 3 passed, across 16 | 3 passed, across 16 |
| lossless | 3 passed against an unspeculated server | 3 passed | 3 passed |

The last column matters because all three lifecycle fixes touch
`_apply_committed_spec_tokens_to_state` and `_propose_model_drafts`, which the
synchronous path runs on every speculative step. Each is a no-op there by
construction (a request always holds a row when a synchronous commit runs, the
step's row is the request's current row, and `skip_req_ids` is never set), and
this run says so rather than the argument.

The re-run exercises the synchronous path only: vLLM refuses asynchronous
scheduling with `custom_class` before the plugin sees the configuration, and
this model does not declare `supports_async_decode` either.

What the suite refuses to do is treat a response as evidence. Every assertion
is a delta on vLLM's own counters across the request:
`spec_decode_num_drafts_total`, `num_draft_tokens_total`,
`num_accepted_tokens_total`, `num_accepted_tokens_per_pos_total{position}`,
`num_preemptions_total` and the `iteration_tokens_total` histogram. The
per-position counters are what distinguish "two of five accepted on every
step" from "five of five on two steps in five", and the configured accept depth
predicts them exactly.

Three specifics worth calling out.

**The batch-size claim is a bound, not an assertion.** One row commits at most
`1+K` tokens in a step, so a step counted above a histogram bucket bound `b`
had more than `b / (1+K)` rows in it. That is the only same-step batch-size
statement this instrumentation supports, and the concurrency test derives it
rather than inferring it from how many requests were posted.

**The capacity tests skip rather than pass** when `num_preemptions_total` did
not move, and the first configuration I tried did not move it: this model
allocates no KV cache, so the pool defaulted to 131072 tokens and no reachable
number of requests could exhaust it. `--num-gpu-blocks-override` cannot fix
that either, because the plugin writes that field itself from the model's
declared budget and an operator's value is silently replaced (worth a separate
look). The model now declares its budget through
`TT_SPEC_MAX_TOKENS_ALL_USERS`, and at 1024 tokens the run reaches 6
preemptions with every request still emitting its whole output.

**Termination is checked inside a committed prefix**, which is the case
speculation creates: `max_tokens` cutting mid-block, a stop token strictly
inside a step's prefix, and the end-of-sequence id, which this model counts
straight through because its `config.json` leaves `eos_token_id` at HF's
`LlamaConfig` default of 2. The server's convention, pinned by these tests, is
that the stopping token is the last id the response carries and the rest of its
step is discarded: a step committing `[0,1,2,3,4,5]` returns `[0,1,2]`.
Streaming and non-streaming are compared at the same boundary, and a cancelled
stream is followed by another request to show the row came back.

`test_lossless_device.py` needs the `fixed` target added in `tt-metal#56646`:
the accept-depth target's output is a function of what was drafted, so it
cannot answer whether speculation is lossless. It runs two servers, one
speculating and one not, and compares whole outputs.

### On the host

```text
PYTHONPATH=ci/host-stubs .venv/bin/python -m pytest tests/ --ignore=tests/tt -q
690 passed in 6.47s      (577 on the base branch)
```

**The transition sequence is driven end to end.** `test_spec_async.py` runs
solo speculation, one resolving verify, three batched ordinary decode steps
with no verify among them, the peer's departure, and speculation resuming, then
checks the solo request's whole output is still its own continuation. Overlap
is asserted as an ordering fact: the first baseline step's readback is held and
the second forward reaches the model anyway. And the barrier is pinned from
both sides, which matters because `check_perform_device_sampling` does not look
at speculation: on a launch that samples on device a verify satisfies every
condition `can_use_steady_decode_fast_path` checks and comes back eligible, so
the explicit registration in `submit_async_decode` is the entire barrier and
the test asserts the predicate's `True` beside the registration's `False`.

Five more mutations applied and caught, including every step staying a verify
(7 tests), a plain step not proposing (1 test synchronous, 2 asynchronous), and
declining a row leaving the last offer standing (1 test).

**The deferred path is driven through the real entry points.**
`tests/spec/test_spec_async.py` goes through `execute_model` and
`sample_tokens` against a target that implements the `read_decode_output` hook
and hands back an output buffer filled only when the test releases it, with the
plugin's `ttnn.event_synchronize` reference patched to wait for that release.
The synthetic events stay inside the test process. Every "before the result
lands" assertion is therefore a real point in the step rather than a sleep.

The output-equality half is that suite's own: the seven draft cases and the
unspeculated reference arm of `tests/spec/test_spec_lossless.py` are imported
rather than restated, and re-run through the asynchronous path, comparing whole
outputs at 25 emitted tokens against both the plain arm and the closed-form
continuation. The instrumentation that proves deferral was exercised: the
request's own history is asserted unchanged at submission and again after
completion, the target's `read_decode_output` call count equals the step count,
and its plain-decode call count stays at zero.

The lifecycle half holds completion open and lets the scheduler act first,
which is where the three defects above came from. It covers completion,
cancellation with the row taken by a new request in the same scheduler step,
preemption followed by a resume that speculates again, compaction moving a
request's row, a forced prefix-cache reset, zero/partial/full acceptance across
three rows with distinct histories in one step, two callers resolving the same
deferred output (one readback, one enqueued commit), each step's proposal
consuming its own verify's hidden handle, and terminal handling of a readback
failure and of a proposal failure. Every ordering is forced by an explicit
release or a `threading.Event`; no test decides its ordering with a sleep.

Seven mutations applied and caught for that half, including the commit writing
to the step's row index (2 tests) and the completed queue being drained after
the apply rather than before (11 tests).

**Losslessness now has a target that can settle it.** The old test compared a
speculated run against a run of the same stand-in with no drafts, which is
still the speculative path, and compared only the shorter arm's length. And
`FakeSpecModel`'s verify returns each draft unchanged wherever it agrees, so
what it "chose" at a candidate position was the draft offered there. It passes
a mutation that shifts every candidate-block position by one.

`tests/spec/deterministic_target.py` chooses its next token by
`(token * 31 + position * 7 + 11) % vocab`, computed identically for an
ordinary decode and for a verify, and never reading `num_valid_drafts`, the
drafts or an accept depth. `tests/spec/test_spec_lossless.py` then compares
whole outputs, at a fixed token count rather than a fixed step count, three
ways: the speculated run, a genuinely unspeculated runner
(`_num_speculative_tokens` of 0, the ordinary host-sampling tail, no accept
walk anywhere), and the closed-form continuation. Seven draft cases: correct,
wrong at the first position, wrong in the middle, wrong at the last, all
wrong, shorter than K, and short and wrong.

**The request lifecycle is exercised for real.** Every other test in
`tests/spec` stands in for `_update_states` by writing
`num_computed_tokens_cpu`. `tests/spec/test_spec_state_transitions.py` drives
the real `_update_states` with real `SchedulerOutput` values through
completion, preemption, resumption, `condense` and row reuse, then runs a real
speculative step across each transition. It asserts what the verify was told,
not only what the runner's maps hold: the accepted count a model picks its
candidate state slot with is invisible in the committed tokens, so the maps
agreeing proves nothing about the wire.

Mutation-checked, eleven mutations:

```text
CAUGHT  the walk accepts the draft that rejected
CAUGHT  committed ids come from the drafts, not the verify
CAUGHT  every candidate-block position shifts by one   (the old test passes this)
CAUGHT  the accepted count is not carried to the next step
CAUGHT  the accepted count is read by row, not by request id
CAUGHT  a first step is told it continued from two tokens
CAUGHT  a finished request's accepted count is never swept
CAUGHT  preemption stops releasing the model's request state
CAUGHT  drafts are reported for requests the runner forgot
CAUGHT  a padding row's committed positions are not restored
SURVIVED the committed block's placeholder boundary moves by one
```

The survivor is equivalent, not untested: `_apply_committed_spec_tokens_to_state`
slices by the accepted count, so the placeholder tail inside the committed
block is never read.

One behaviour changed with them: `take_draft_token_ids` drops requests the
runner no longer holds. A request can finish in the step that proposed for it;
the scheduler tolerates a draft for a finished request
(`requests.get` then an `is_finished` skip), but reporting one for a request
the runner has forgotten is not the runner's to do.

`tests/spec/test_spec_greedy_e2e.py` drives the drafter through the real
submission path: the propose call and its rows, the drafts reaching
`take_draft_token_ids`, the hidden handle's identity from verify to drafter,
two rows continuing from their own committed tails, and every step committing
`1+K` with nothing draftless. `tests/spec/test_spec_admission.py` covers the
method's admission, its pinned `model` path, and the two declarations a model
needs. `tests/spec/test_spec_descriptors.py` reaches both new methods off a
real instance, because a `SimpleNamespace` hides a missing or spurious
`@staticmethod`.

One defect was found that way before the first commit: the test harness bound
`_committed_positions`, a static method, which shifted every argument by one.

An external review of the first commit found four more, all of them reachable
by a conformant model and none reachable by the paired test model, which is why
the device run did not show them: the padding-row positions and the narrow-step
width described above, the missing dtype check, and a hidden-handle regression
test that copied the handle in the harness and so would have stayed green had
`_forward_with_model_input` stopped copying it. Each fix is mutation-checked,
including both halves of that copy, and the per-row continuation test now gives
its two rows different prompts and different accepted counts, so a row-mixing
defect fails it (verified by mutating the runner to read row 0 for every row).

## Checklist


viktorpusTT:
Some performance findings from measurements with the dummy model:
* The plugin's cost of speculation is around 0.6 ms for B=1 and does not change with model speed.
* Each additional drafted token costs about 0.032 ms per output token at one row and 0.044 ms at eight rows
* Thirty-two times the rows costs three times the penalty, not thirty-two times.
* Async decode does not improve TPOT.
* The slower the model, the closer speculation gets to its ideal. At a fifty-millisecond model duration, accepting all drafts cuts TPOT by 5.72 times.
* On a model whose decode step costs 5 ms or more, speculation pays for itself as soon as the mean committed width exceeds about 1.1 to 1.2.

PR #134, opened 2026-09-17
fix: skip persistent state slots for stateless models

nkapreTT:
## Problem

Stateless paged models use scheduler page tables as their durable request identity, but the plugin currently also retains every prior request in a fixed 32-entry device-state-slot map. A new B32 scheduler wave can therefore fail even when the scheduler correctly admits it. In Shield run `35192192215`, GPT-OSS completed GPQA, started MMLU, then crashed while admitting 31 prefills because all 32 slots were still held by requests from the prior wave.

## Change

- normalize the generic `requires_persistent_request_state_slots` model capability at platform admission;
- retain the historical persistent-slot behavior by default;
- use row-local slots for explicitly stateless prefill submissions and no decode remap;
- leave recurrent/stateful models on the existing request-owned slot path;
- cover the exact 32-held → 31-new regression and the stateful fail-closed path.

There is no model-name check, host fallback, or generated-model bypass.

## Evidence

- Exact failing Shield image reproduced the pre-fix exception with 32 held slots and 31 new prefills.
- The patched source mounted read-only into that same image returned row-local slots `0..30` for the explicit stateless capability.
- The same patched image still raised slot exhaustion for the persistent-state capability.
- Focused host tests: `129 passed`.
- Full host suite: `520 passed`.
- Ruff and `git diff --check`: clean.

Silicon requalification remains required before this becomes a production pin.


viktorpusTT:
This seems to be a bug that affects all models, regardless whether they hold state or not (should be fixed generally). Sometimes the info about the finished requests did not propagate from Scheduler to Runner, which resulted in slots falsely occupied. Here is a proposed fix: https://github.com/tenstorrent/vllm-tt-plugin/pull/135

viktorpusTT:
@nkapreTT #135 merged, can you try your test?

